PR #141, opened 2018-07-18
Fix backend tests

zjmiller:
Just a few fixes that I came across when trying to run the backend tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Mosaic 
+# Mosaic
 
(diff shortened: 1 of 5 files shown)


zjmiller:
For reference, you can see how the "use_env_variable" field in Sequelize's `config.json` works by looking at this part of the Sequelize source code.

https://github.com/sequelize/cli/blob/ff719146589d28c198493e5b57b1aaa6a3cd0a73/lib/tasks/db.js#L191

zjmiller:
Sounds good!

PR #142, opened 2018-07-19
Fix homepage ordering of workspaces

zjmiller:
Previously workspaces were being ordered by _.sortBy on their createdAt
property. But createdAt was just a string, like
"Wed Jul 18 2018 21:42:56 GMT+0000 (UTC)". So these strings were being
compared alphabetically, and this order was determining the order in
which the workspaces appeared on the homepage. I made sure it was the
date itself that was being used to create the sort order. I also put
this sort function in its own file.


zjmiller:
Makes sense!

Issue #140, opened 2018-07-18
Make pointer expansion less fiddly


stuhlmueller:
E.g. move the button closer to the pointer text, or (for read-only pointers) allow the user to toggle expansion status by clicking on the text directly.

PR #143, opened 2018-07-19
Click read-only pointers directly to toggle expansion

zjmiller:
See https://github.com/oughtinc/mosaic/issues/140

I'm still getting a feel for the pointer system. Not 100% confident this is the best way to implement this feature. As I understand more of the big picture I might refactor.

### DIFF
--- a/client/src/lib/slate-pointers/PointerImportNode.tsx
+++ b/client/src/lib/slate-pointers/PointerImportNode.tsx
@@ -4,7 +4,7 @@ import { connect } from "react-redux";
 import styled from "styled-components";
 import { ShowExpandedPointer } from "./ShowExpandedPointer";
 import { propsToPointerDetails } from "./helpers";
-import { changePointerReference } from '../../modules/blockEditor/actions';
+import { changePointerReference } from "../../modules/blockEditor/actions";
 
 const RemovedPointer = styled.span`
   background-color: rgba(252, 86, 86, 0.66);
--- a/client/src/lib/slate-pointers/PointerImportNode.tsx
+++ b/client/src/lib/slate-pointers/PointerImportNode.tsx
@@ -68,12 +68,12 @@ class PointerImportNodePresentational extends React.Component<any, any> {
     }
   };
 
-  public handleClickingClosedPointer = (e, pointerId) => {
+  public handleClosedPointerClick = (e: Event, pointerId: string) => {
     this.props.openClosedPointer(pointerId);
-    e.stopPropagation();
+    e.stopPropagation( );
   }
 
-  public handleClickingOpenPointer = (e, pointerId) => {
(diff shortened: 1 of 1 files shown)


stuhlmueller:
This looks like a good start, but doesn't do the right thing for nested pointers yet:

nested-pointers

zjmiller:
Ah good catch.

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -66,6 +68,16 @@ export class PointerImportNode extends React.Component<any, any> {
     }
   };
 
+  public handleClickingClosedPointer = (e, pointerId) => {
```
stuhlmueller: Use type annotations for function parameters? There are many places that aren't using them at the moment, and we don't need to convert everything now, but I think for newly added code, it makes sense to add annotations.

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -66,6 +68,16 @@ export class PointerImportNode extends React.Component<any, any> {
...
+    this.props.openClosedPointer(pointerId);
+    e.stopPropagation();
+  }
+
+  public handleClickingOpenPointer = (e, pointerId) => {
```
stuhlmueller: I'd have used `handleOpenPointerClick` and `handleClosedPointerClick` based on the analogy with `handleClick`, which is much more common than `handleClicking`.

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -94,6 +106,7 @@ export class PointerImportNode extends React.Component<any, any> {
     if (!isOpen) {
       return (
         <ClosedPointerImport
+          onClick={e => this.handleClickingClosedPointer(e, this.props.nodeAsJson.data.internalReferenceId)}
```
stuhlmueller: Could make sense to factor out to `const pointerId = this.props.nodeAsJson.data.internalReferenceId` since it's used here and below.

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -117,3 +133,17 @@ export class PointerImportNode extends React.Component<any, any> {
     }
   }
 }
+
+const mapDispatchToProps = dispatch => ({
```
stuhlmueller: Add typescript annotations here and below?

client/src/lib/slate-pointers/PointerImportNode.tsx
zjmiller: Sounds good. I've used Flow more than Typescript but I don't think it will take me long to get the hang of it. I'll do some reading up on my own time.

PR #144, opened 2018-07-24
Optimize backend tree loading

zjmiller:


zjmiller:
I connected [https://mosaic-prod-pr-144.herokuapp.com/](https://mosaic-prod-pr-144.herokuapp.com/) to the production database so that we can test the differences in loading time.

client/src/components/WorkspaceCard/index.tsx
```diff
@@ -80,36 +90,43 @@ export class WorkspaceCardPresentational extends React.PureComponent<
...
+            .map(node => ({...node, readOnly: !editable }))
+            .value()
+          :
+          []
+        );
 
     if (!workspace) {
-      return <div>Loading...</div>;
+      return (
+        <div>
```
stuhlmueller: Could factor out this div as a stateless component (in same file).

client/src/components/WorkspaceCard/index.tsx
```diff
@@ -41,13 +44,20 @@ interface ConnectedPointerType {
 interface WorkspaceType {
   blocks: any[];
   childWorkspaceOrder: string[];
-  connectedPointers: any;
+  connectedPointersOfSubtree: ConnectedPointerType[];
   id: string;
 }
 
 interface WorkspaceCardProps {
+  isChild: boolean;
```
stuhlmueller: `isChild` doesn't feel quite right as a name. I don't have a great proposal myself, but think the relevant concept is "is this workspace the root node of the subtree we're showing?". (Feel free to leave as is if you can't think of something better.)

server/lib/models/pointer.ts
```diff
@@ -85,6 +115,24 @@ const PointerModel = (sequelize, DataTypes) => {
     return pointerIds;
   };
 
+  Pointer.prototype.newDirectContainedPointerIds = async function(pointersSoFar) {
+    const value = await this.value;
+    if (!value) {
```
stuhlmueller: When can this happen?

server/lib/models/workspace.ts
```diff
@@ -237,6 +257,32 @@ const WorkspaceModel = (
...
+    const blocks = await this.getBlocks();
+
+    for (const block of blocks) {
+      const blockPointersToAdd = await block.newConnectedPointers(pointersSoFar);
+      connectedPointersOfSubtree = connectedPointersOfSubtree.concat(blockPointersToAdd);
+      pointersSoFar = pointersSoFar.concat(blockPointersToAdd);
+    }
+
+    for (let i = 0; i < this.childWorkspaceOrder.length; i++) {
+      const childWorkspaceId = this.childWorkspaceOrder[i];
```
stuhlmueller: Why not use `for ... of ...` here like above?

server/lib/models/pointer.ts
zjmiller: I don't know. There are a lot of guards like this throughout the codebase where it's unclear how the situation guarded against can arise. I think it would take quite a bit of detective work to figure out if they're still needed. Seems faster to just minimally modify the `directContainedPointerIds` method.

PR #145, opened 2018-07-24
Show names of exported pointers

### LINKED ISSUE #81: Show names of exported pointers inline

zjmiller:
I removed a lot of old code. Everything seems good but please give this PR a little extra manual testing if you have the time!

Closes #81


stuhlmueller:
Nested pointers don't look how I'd expect them to look:

nested-pointer-names

What I expected (with spaces for clarity):

```
$3[ T $4[ hi ] s ]  is a test
```

zjmiller:
I'm getting that same behavior on the current version of the app. You just can't tell in admin mode because there aren't brackets and names. But if you perform the same test on the current version of the app, and then go into the tree view, you'll see what I mean. I checked out old versions of the app and the same thing happened, so the issue wasn't recently introduced. I'll look into it and plan to push the fix to this branch.

stuhlmueller:
You're right. After a bit more experimentation, it looks like it works when you create the inner pointer first, but doesn't work when you create the outer one first. In the past, I must have always created the inner pointer first, and so didn't notice.

stuhlmueller:
Another thing to be on the lookout for is usability issues that arise when navigating in the text box that includes the extra annotations. I noticed that when I used left/right keys to move the cursor, it didn't always let me move past the brackets + $ signs.

zjmiller:
Noticed that removing nested pointers doesn't work as expected. Working on that now!

zjmiller:
I should note this last commit, the vey big pointer style commit, is not really backwards compatible. Pointers created before this change won't look or behave quite right as far as the UI goes.

zjmiller:
There are only four remaining UI issues with this change that I'm aware of:

1. With nested imported pointers, the nested pointer has a little more padding before it than after it.
2. When an exported pointer starts off as not nested, but then gets nested within another pointer, there are some slight spacing irregularities.
3. Conversely, when an exported pointer starts off as nested, but then is no longer nested because its parent pointer is removed, there are some slight spacing irregularities.
4. If you backspace across a pointer boundary, you can accidentally remove some spacing that really shouldn't be able to be removed. This will cause some spacing and cursor/selection issues.

stuhlmueller:
Of these, 4 sounds most important. For 2 and 3, the issue itself doesn't sound that important, but it would be good to understand the underlying cause.

stuhlmueller:
Nice to see that mouse dragging can now select across pointers!

With ctrl-d (forward delete), it's currently still possible to delete the spaces around the brackets and mess up the cursor navigation:

ctrl-d

I wonder if any approach that inserts text spaces can be robust. I think we shouldn't be in the situation where we try to patch ways to delete the spaces one by one, but rather should have some guarantee that if there's a pointer, it's definitely rendered correctly.

zjmiller:
I agree that ideally we wouldn't have to insert spaces into the content of the pointer, both because it seems like a hack and because we then worry about having those spaces get deleted or moved. But in this context I'm starting to think it is unavoidable. The problem is that when we have an exported pointer, we sometimes want to insert text at the start/end of the pointer, and we sometimes want to insert text directly before/after the pointer. So we need two cursor positions to correspond to these two different ways of inserting text. But without extra characters for the cursor to move over, we only have one cursor position, and we can't distinguish all of these possibilities.

(As an aside: Using block elements, instead of inline elements, actually gives us these two cursor positions. For example, if the cursor is directly after the last character of a block element, pressing the right arrow will move it to directly before the first character of the next block element. So, visually, it seems like the cursor isn't moving at all, but this is kinda what we want in this case. The trouble is that for Mosaic we need inline elements if we want our pointers to soft wrap in the right way.)

So we need these extra characters padding our pointers. Ideally we could have Slate just always render them on top of the pointer content (using something like Slate's custom nodes), instead of putting them directly in the pointer content. In theory in this case they couldn't be moved or deleted, because you're basically telling Slate to _always_ render an exported pointer with spaces at the start/end and before/after. It's just part of the rendering instructions, not part of the pointer content. But characters that are added to a Slate editor in this way (as part of the rendering instructions and not as part of the actual editable content) are invisible to Slate's internal system for tracking cursor position, which means you don't solve any of the problems listed above (there ends up being just one cursor position where you need two) and you actually introduce new problems (arrow key navigation stops working as expected).

There is actually a Slate plugin called [slate-sticky-inlines](https://github.com/YurkaninRyan/slate-sticky-inlines) that is supposed to allow editing at the edge of an inline. If this worked it would get us some way toward a solution. But the plugin is buggy and doesn't work with nested inlines, among other drawbacks, so is a non-starter.

zjmiller:
One thing that would be helpful is if we can find an example of a web-based text editor that exhibits some or all of the behavior we want. Mainly, we're looking for a text editor where there's some inline element and where a user can naturally add characters both before and after it, and at the start and end of it. Ideally, this is a Slate editor. But even an example of a non-Slate editor that can do this would be useful, even though the way it does it might turn out to be hard/impossible to implement with Slate.

zjmiller:
For now my plan is to take a middle path, where I can the current approach of inserting spaces into the pointer content, but I'll add a normalization process in the editor component to guarantee that the relevant text nodes have the requires spaces. This approach will make it impossible to get the editor into a state where the requires spaces are absent. And this approach will mean we don't have to worry about individual patches for every possible way of deleting/moving a character.

stuhlmueller:
Thanks for the detailed explanation. I think your compromise makes sense.

Given that we're going to add pointer creation using square brackets, would it work for the "extra" characters needed to allow insertion before/after pointer start to just be the square bracket? Or is there a reason why using spaces and rendering the square bracket separately is better?

stuhlmueller:
While I was testing this PR, I came across issue #147 (which probably wasn't introduced here, so feel free to postpone fixing it to another PR).

stuhlmueller:
For exported pointers in sub-questions, we currently show name `$0` for each, see this gratuitous animation:

subquestion-pointers-1

After submission, we see the real names:

subquestion-pointers-2

I think our current setup makes it difficult to show the real names that pointers will have after submission, and to allow the user to refer to those names. But it could be worth indicating in some other way that the pointers are still in production and can't be referred to yet.

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -176,7 +246,49 @@ export class BlockEditorEditingPresentational extends React.Component<
     }
   };
 
-  private onKeyDown = event => {
+  private onKeyDown = (event, change) => {
```
stuhlmueller: Add types?

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -13,6 +13,19 @@ import { valueToDatabaseJSON } from "../../lib/slateParser";
 import { exportSelection } from "../../modules/blockEditor/actions";
 import * as _ from "lodash";
 import { UPDATE_BLOCKS } from "../../graphqlQueries";
+import { SPACER } from "../../lib/slate-pointers/exportedPointerSpacer";
+
+interface Change {
+  insertTextByKey: any;
```
stuhlmueller: Probably worth turning into a function type, even if the arguments are `any`.

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -123,6 +136,63 @@ export class BlockEditorEditingPresentational extends React.Component<
     }
   }
 
+  public normalizeChangeWrtTextNodeSpacing(c: Change) {
```
stuhlmueller: `BlockEditorEditing.tsx` is getting pretty big. Consider abstracting out parts. Text normalization seems like a natural candidate.

client/src/components/BlockEditor/BlockEditorEditing.tsx
stuhlmueller: Also, `onKeyDown` is a bit difficult to reason about because it's a long sequence of conditionals and actions (that are themselves kind of difficult to think about for someone like me who doesn't know Slate well). I wonder if there are some abstractions that could make the function easier to read, and that could make it more obvious that it's behaving correctly.

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -202,6 +318,52 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+    // the onClick handler didn't seem to have access to the new cursor position
+    // so this logic lives here instead
+
+    const value = c.value;
+    const selection = value.selection;
+
+    const anchorKey = selection.anchorKey;
+    const focusKey = selection.focusKey;
+    const anchorOffset = selection.anchorOffset;
+    const focusOffset = selection.focusOffset;
```
stuhlmueller: Can simplify: `const { anchorKey, focusKey, anchorOffset, focusOffset } = selection;`

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -202,6 +318,52 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+
+    const selectionIsExpanded = (anchorKey !== focusKey) || (anchorOffset !== focusOffset);
+
+    // if selection is expanded, then we don't do this
+    // this allows mouse dragging to select across pointers
```
stuhlmueller: The code below is somewhat difficult to read without knowing Slate. I wonder if there are abstractions that would help. It's also plausible that the behavior in the `!selectionIsExpanded` case should be factored out.

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
```diff
@@ -57,29 +57,40 @@ export class ShowExpandedPointer extends React.Component<any, any> {
...
-              <InlineNode
-                node={node}
-                key={index}
-                blockEditor={this.props.blockEditor}
-                availablePointers={this.props.availablePointers}
-                pointerId={this.props.exportingPointer.pointerId}
-                onMouseOver={this.props.onMouseOverPointerImport}
-                isHoverable={this.props.isHoverable}
-              />
+              <span key={index} style={{ marginLeft: "4px", marginRight: "6px" }}>
```
stuhlmueller: Use styled-components for consistency?

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
```diff
@@ -57,29 +57,40 @@ export class ShowExpandedPointer extends React.Component<any, any> {
   }
 
   public render() {
+    const isNestedAsExportingPointer = this.props.availablePointers.some(pointer => {
```
stuhlmueller: This seems like it might become expensive for larger trees. For each expanded pointer, each time we render, we traverse (up to) all available pointers (which can be very many given our setup), and for each of them, we traverse (up to) all nodes. Is there some other way to achieve what this is doing? Or is there reason to think that it's not going to affect rendering speed much?

client/src/lib/slate-pointers/index.tsx
```diff
@@ -11,7 +11,12 @@ function SlatePointers(options: any = {}) {
       }
 
       const selection = window.getSelection();
-      const range = selection.getRangeAt(0);
+      let range;
+      try {
+        range = selection.getRangeAt(0);
+      } catch (e) {
```
stuhlmueller: Do we need to catch all exceptions here, or can we just catch a specific type and reraise otherwise?

client/src/modules/blockEditor/actions.ts
```diff
@@ -62,11 +63,63 @@ export const exportSelection = () => {
     const { hoveredItem } = blockEditor;
 
     const block = blocks.blocks.find(b => b.id === hoveredItem.blockId);
+
     const uuid = uuidv1();
     if (block) {
+      const fragment = block.value.fragment;
+      const topLevelNodes = fragment.nodes.toJSON()[0].nodes;
+
+      const onlyOneNodeThatIsPointerExport = nodes => {
```
stuhlmueller: Move the two functions and `spaceTextNode` somewhere closer to top-level so that they aren't redefined every time the export selection handler is called?

client/src/modules/blockEditor/actions.ts
```diff
@@ -62,11 +63,63 @@ export const exportSelection = () => {
...
+
+      let nodes = topLevelNodes;
+
+      while (
+        onlyOneNodeThatIsPointerExport(nodes)
+        ||
+        twoNodesFirstExportSecondEmptyString(nodes)
+      ) {
+          nodes = nodes[0].nodes;
+        }
```
stuhlmueller: What's the intention/logic behind the six lines above?

client/src/modules/blockEditor/actions.ts
```diff
@@ -114,11 +167,48 @@ export const removeExportOfSelection = () => {
...
+      const nodeToRemove = matchingNodes[0];
+
+      const ancestorNodes = block.value.document.getAncestors(nodeToRemove.key);
+
+      const isNested = ancestorNodes.find(ancestor => ancestor.type === "pointerExport");
+
+      let change;
+      change = block.value.change();
+
+      // delete spacer at start and end
```
stuhlmueller: The spacer deletion here and below feels brittle. Can this be handled using the normalization process somehow?

zjmiller:
> I should note this last commit, the vey big pointer style commit, is not really backwards compatible. Pointers created before this change won't look or behave quite right as far as the UI goes.

This should not be an issue now that the editor component automatically guarantees that the relevant nodes have the required spacer characters. So the changes shouldn't result in any loss of functionality with the old trees.

There is one issue, though, which is that old trees will still have a bug that the new trees don't. As you noticed above, when you export a pointer inside a pointer, old trees will still exhibit the weird "splitting" behavior:

image

New trees don't have this happen because their nodes are structured somewhat differently. I wasn't planning on restructuring the old trees so that they possess the new bug-free structure. Though this is something I could look into if we really want to keep using and editing the old trees. I'm not sure it's worth it, though, since it is only relevant for old trees.

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/a245a2c623007051c9b6d4c0a0bd4c4e83fc26da

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/c640cee425a03d2a40bae57bec6470e8fe336b8b

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: I cleaned it up pretty significantly here: https://github.com/oughtinc/mosaic/pull/145/commits/d6197a53bfccb542974d2501f05914dc084db102. Still need to add types and think about best way to add abstraction.

client/src/modules/blockEditor/actions.ts
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/ab7abae57db354aba6cb665c102a4f813b9353e4

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
zjmiller: I think styled-components make sense when you're reusing the styles, or when you need something like a selector or pseudo-selector, or when you're using a newish feature and want automatic vendor prefixing, but otherwise I find it makes files harder to parse. When the style information is more closely coupled to the markup structure, I have a better feel for the CSS structure (what styles are nested in what) and an easier time making quick changes. I think this is especially true when the styling is just a few pixels of padding or margin.

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
zjmiller: I agree it doesn't look too scalable. Given the current approach to pointers, we actually don't really need it. So for now we can just take it out entirely. If, in the future, in this component, we want to know whether a pointer is nested, we can revisit the issue of how to most efficiently to determine this.

zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/1677f11d0f392b823d4030f19d6db22a9de25816

client/src/lib/slate-pointers/index.tsx
zjmiller: Good question. This error was occurring too infrequently and unpredictably for me to figure out what was going on. Also, a lot of code has changed since that happened. I think we should probably just stop swallowing errors here and see if they still occur. If they do hopefully we can figure out why, but there's a chance that the newest approach avoids them entirely.

client/src/components/BlockEditor/normalizeChange.js
```diff
@@ -0,0 +1,53 @@
+import { SPACER } from "../../lib/slate-pointers/exportedPointerSpacer";
+
+export function normalizeChangeWrtTextNodeSpacing(c: Change) {
```
stuhlmueller: Do you need to import the `Change` type?

stuhlmueller:
> I wasn't planning on restructuring the old trees so that they possess the new bug-free structure. Though this is something I could look into if we really want to keep using and editing the old trees. I'm not sure it's worth it, though, since it is only relevant for old trees.

Agreed, not worth it.

client/src/lib/slate-pointers/index.tsx
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/d22a45803ee9687ae1ebb5e019247f7cf249e9e4

client/src/modules/blockEditor/actions.ts
zjmiller: Basically trying to get to the inner-nested nodes of a fragment. I added an explanatory comment here: https://github.com/oughtinc/mosaic/pull/145/commits/529628930bf22b2a36fa72d37e6d81dfa3d62c0c

I'll think if there might be a simpler way to approach this.

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: OK added types here https://github.com/oughtinc/mosaic/pull/145/commits/53d0c601af59cb7397814b4cc94ab4fcece3fd4d

After this branch is merged, I'll make a branch where I just clean up all the types. At least put them in a dedicated folder/file so we avoid duplicate work in the future.

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: Not exactly what you were suggesting, but I factored out the logic for determining whether a selection is expanded or not: https://github.com/oughtinc/mosaic/pull/145/commits/73b23db2e2fd114e3387e7516f6b340cf8826eb9

client/src/modules/blockEditor/actions.ts
zjmiller: This code checks to see whether the spacer character exists at the start and end of the text of the pointer to be removed, and before and after the pointer to be removed. If the spacer character exists at any of those positions, it's deleted. 

If any of those nodes don't exist, or if the character at the relevant position isn't the spacer character, then the deletion doesn't occur. So this code might be brittle in some ways, but it abstracts over the spacer character, and it's cautious about when it actually deletes a character. I'm not too worried about it.

That said, I'd be happy to find a more robust way of doing this.

One thing we could do is expand the normalization process to make sure that *both* **(1)** the spacer character (currently thin space) occurs at the start/end of, and before/after, every pointer and **(2)** the spacer character *only* occurs in those positions (so delete all of the spacer characters not in those positions). 

Currently, we only do **(1)**, and allow users to insert additional spacer characters in their editors if they wish to. Pursuing **(2)** would take all of this "pointer removal cleanup" and include it in the normalization process, but it would mean users can't use thin spaces at their discretion (which might be fine).

client/src/components/BlockEditor/isSelectionExpanded.js
```diff
@@ -0,0 +1,21 @@
+export function isSelectionExpanded(selection) {
```
stuhlmueller: Why can't we use Slate's [`isExpanded`](https://docs.slatejs.org/slate-core/range#isexpanded)? Is selection not a range?

client/src/modules/blockEditor/actions.ts
stuhlmueller: How can I see that the cases are the only relevant ones, i.e. how do I know that things like [empty string, export] or [export, empty string, empty string] aren't possible?

client/src/modules/blockEditor/actions.ts
```diff
@@ -114,11 +174,51 @@ export const removeExportOfSelection = () => {
...
-        .change()
-        .unwrapInlineByKey(matchingNodes[0].key, {
-          data: { pointerId: hoveredItem.id }
-        });
+      const nodeToRemove = matchingNodes[0];
+
+      let change;
+      change = block.value.change();
+
+      // delete spacer at start
```
stuhlmueller: I think I already asked this somewhere (in that case, just point me to your previous response): Why do we need to manually delete the spacers below? Why can't we just remove the export and then rely on normalization to fix the spacing?

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
stuhlmueller: Sounds reasonable.

client/src/components/BlockEditor/normalizeChange.js
```diff
@@ -0,0 +1,54 @@
+import { SPACER } from "../../lib/slate-pointers/exportedPointerSpacer";
+import { Change, TextNode } from './types';
+
+export function normalizeChangeWrtTextNodeSpacing(c: Change) {
```
stuhlmueller: This is a pretty unwieldy function name. How about `normalizeExportSpacing`?

client/src/components/BlockEditor/isSelectionExpanded.js
zjmiller: Oh yeah we could use that. Not sure how I missed that. I'll switch it.

client/src/modules/blockEditor/actions.ts
zjmiller: Just because normalization doesn't remove characters. We could change it so that it does, but then users couldn't insert thin spaces (or whatever the spacer character is) in other positions.

client/src/components/BlockEditor/normalizeChange.js
zjmiller: Sure.

client/src/modules/blockEditor/actions.ts
stuhlmueller: Is there some way to attach metadata to the spacer characters so that it's distinguished from normal ones?

Having normalization take care of the entire export space formatting seems like a substantial simplification to me.

zjmiller: With Slate, as far as I can figure out, it's actually really complicated to add metadata like that. Basically, you can add a mark, which is like formatting, so it is similar to bold or italics. In the case of metadata, the mark just wouldn't correspond to any visual difference. 

But then you have to add a lot of logic to make sure new text you insert doesn't pick up the mark (think of how you can put your cursor next to bold text, start typing, and insert more bold text).

I would be OK with just saying that users can't use the spacer character, especially when the spacer character is "non-breaking thin space". I could also put a comment in the file that determines the spacer character, warning developers that the spacer character is not something the user can manipulate (so don't set it to normal space, for example). If we did this, then we could expand the normalization process in the way you're thinking.

stuhlmueller: Sounds good. To be extra-safe, we could programmatically prevent users from inserting the spacer character. That's probably overkill for now, but you can make an issue for it.

client/src/modules/blockEditor/actions.ts
zjmiller: Typically, when there is more than just the exported pointer, you actually want to include the exported pointer in the content of what you're exporting (thus creating a nested export). The [export, empty string] that I'm currently looking for is an edge case when the text you've selected goes all the way to the end of the pointer. For example, suppose you've exported the content `AAABBB`, and now you want to create a nested pointer that just encompasses the `BBB`. When the expanded selection encompasses `BBB`, the end of the selection is actually at the start of the next node. If were selecting `AAA`, the start of the selection is not at the end of the previous node. So with cursor selection, there is an asymmetry between the start and end.

This actually interacts with the spacer characters in a way that I'm currently addressing.

client/src/components/BlockEditor/isSelectionExpanded.js
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/42d31e63e082390a02e4978d04a662b06905112e

client/src/components/BlockEditor/normalizeChange.js
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/f8187e31d855b923d4aed229263bed31ab2cf8b4

client/src/modules/blockEditor/actions.ts
zjmiller: OK the logic here is a bit different now: https://github.com/oughtinc/mosaic/pull/145/commits/b16d2070776d6cdc40e3942a503b28fca30e4f3a

What I'm doing is first checking to see whether we're already in a pointer when we're trying to export. If we are, then I check to see whether the selection is up against an edge. If it is, then I move it over so that it's no longer up against an edge. This means I can now ignore the possibility where we are worried about [export, empty string], since that is a possibility that only arises when the selection extends to the end of the pointer.

client/src/modules/blockEditor/actions.ts
zjmiller: https://github.com/oughtinc/mosaic/pull/145/commits/01608fdbc36018c097e3f22d9c7c172939c703fc

client/src/modules/blockEditor/normalizeAfterRemoval.ts
```diff
@@ -0,0 +1,20 @@
+import { SPACER } from "../../lib/slate-pointers/exportedPointerSpacer";
+import { Change } from "../../components/BlockEditor/types";
+import { normalizeExportSpacing } from "../../components/BlockEditor/normalizeChange";
+
+export function normalizeAfterRemoval(c: Change) {
```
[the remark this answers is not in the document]
zjmiller: I'm worried the removal process is a bit costly when the editor content is large or complicated. Also, I'm assuming that pointer removal will be a relatively rare user action, so I'm hesitant to trigger the process every time a user inserts or deletes a character (i.e., many times per second).

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -176,7 +186,31 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+    if (isMovingRight) {
+      valueAfterSimulatedChange = valueAfterSimulatedChange.change().move(1).value;
+    }
+
+    if (isCursorInPotentiallyProblematicPosition(valueAfterSimulatedChange)) {
+      const correctedChange = handleMovingCursor(change, valueAfterSimulatedChange, isMovingLeft, isMovingRight);
+      this.props.updateBlock({ id: this.props.block.id, value: correctedChange.value, pointerChanged: false });
+      event.preventDefault();
+      return false;
+    }
```
OAGr: In the very least I think the above should mostly be in a separate function or similar. I kind of get what it's doing, but think it could be further cleaned up.

OAGr: This also seems like stuff I could imagine being put into a Slate library or similar. Most of the other contents of this file have to do with things really orthogonal to the editing process itself; like general component mounting, updating the database, and updating the redux stores. Much of the above seems much more text-editor specific.

client/src/components/BlockEditor/types.tsx
```diff
@@ -4,3 +4,25 @@ export enum MutationStatus {
...
+  key: string;
+  object: string;
+  leaves: any[];
+  text: string;
+}
+
+export interface Value {
+  document: any;
+  selection: any;
+}
```
OAGr: Kudos for adding the types here. Do you think we could take some of these from Definitely Typed instead, to keep it up to date with the Slate lib?

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/slate

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -201,7 +240,31 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+
+    const value = c.value;
+    const selection = value.selection;
+    const selectionIsExpanded = selection.isExpanded;
+
+    // if selection is expanded, then we don't do this
+    // this allows mouse dragging to select across pointers
+    if (!selectionIsExpanded && isCursorInPotentiallyProblematicPosition(value)) {
+      const correctedChange = handleStationaryCursor(c);
+      this.props.updateBlock({ id: this.props.block.id, value: correctedChange.value, pointerChanged: false });
```
OAGr: The first thing that onChange() does is call ``updateBlock``. So in this case updateBlock would be called twice. Is there any way we could compress this? Could the value in line 268 just be the ``correctedChange.value`` in this case?

client/src/lib/slate-pointers/PointerExportMark.tsx
```diff
@@ -92,24 +35,57 @@ export class PointerExportMark extends React.Component<any, any> {
...
+          font: "800 1.2em sans-serif",
+        },
+
+        ":after": {
+          color: "rgba(12, 165, 0, 0.63)",
+          content: `"]"`,
+          font: "800 1.2em sans-serif",
+        },
+      },
+    });
```
OAGr: Happy that this :before / :after use works. I would appreciate it if this could be DRYed up a bit. I'm also not totally sure this can't be done with styled-components, with a bit of trickery, I may mess with that later.

client/src/lib/slate-pointers/ShowExpandedPointer.tsx
OAGr: I think I'm pretty ok with this; in cases where there are only a few position adjustments, splitting it out can be a bit of a pain.

client/src/utils/slate/handleStationaryCursor.ts
```diff
@@ -0,0 +1,25 @@
+import { Change } from "../../components/BlockEditor/types";
+
+export function handleStationaryCursor(change: Change) {
```
OAGr: I think the name of this could be more clear. "handleStationaryCursor" really doesn't tell me much about what's going on; from the above use it seems to be doing some DOM manipulation without really returning anything important. Does this need to return "change", if in the one case it's being used, it doesn't accept that return? (Just to be more obvious what's going on.)

Maybe "adjustCursorIfAtEdge"

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -201,7 +210,29 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+    // the user from using a click top get the text cursor in off limit spacing
+    // the onClick handler didn't seem to have access to the new cursor position
+    // so this logic lives here instead
+
+    const value = c.value;
+    const selectionIsExpanded = value.selection.isExpanded;
+
+    // if selection is expanded, then we don't do this
+    // this allows mouse dragging to select across pointers
+    if (!selectionIsExpanded && isCursorInPotentiallyProblematicPosition(value)) {
```
OAGr: Can these two checks be in the file for ``handleStationaryCursor``, given that it already has other checks?

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -201,7 +210,29 @@ export class BlockEditorEditingPresentational extends React.Component<
     }
   };
 
-  private onChangeCallback = (c: any) => {
+  private onChangeCallback = (c: Change) => {
+    // first check to see if document changed, which we can do with !== b/c
+    // Slate uses immutable objects, if it has changed then normalize wrt
+    // pointer spacing
```
OAGr: I appreciate the comments, but one thing that I would have done first is just to write this as: 
```
const documentHasChanged = c.value.document !== this.props.block.value.document
c = documentHasChanged ? normalizeExportSpacing(c) : c
```
or similar

client/src/utils/slate/handleMovingCursor.ts
```diff
@@ -0,0 +1,26 @@
+import { Change, Value } from "../../components/BlockEditor/types";
+
+export function handleMovingCursor(change: Change, value: Value, isMovingLeft: boolean, isMovingRight: boolean) {
```
OAGr: If this function has 4 arguments, I'd prefer that it have a hash input or similar. Typically I prefer hash inputs for anything over 2 args.

client/src/modules/blockEditor/actions.ts
```diff
@@ -56,17 +59,77 @@ export const changePointerReference = ({ id, reference }) => {
...
+        ||
+        twoNodesFirstExportSecondEmptyString(initialTopLevelNodes);
+
+      const change = value.change();
+
+      // if we are exporting a nested pointer, guarantee that the selection is
+      // not extending to an edge, because selections on the edge of an inline
+      // pick up what comes after
+      if (isNestedInPointerExport) {
+        moveSelectionAwayFromPointerEdge(change);
```
OAGr: Does this function mutate change? If so, is there any way you could write it into an immutable function, that returns a different change object instead?

client/src/lib/slate-pointers/exportedPointerSpacer.ts
```diff
@@ -0,0 +1 @@
+export const SPACER = "\u202f"; // non-breaking thin space
```
OAGr: I'm not a big fan of the word SPACER. Given how important it is, I'd prefer something descriptive to something short. Maybe something like, "POINTER_EDGE_SPACE"?

client/src/modules/blockEditor/actions.ts
```diff
@@ -114,11 +177,24 @@ export const removeExportOfSelection = () => {
...
+
+      if (isNested) {
+        change
+          .unwrapInlineByKey(nodeToRemove.key,  { data: { pointerId: hoveredItem.id } })
+          .removeNodeByKey(nodeToRemove.key);
+      } else {
+        change.unwrapInlineByKey(nodeToRemove.key,  { data: { pointerId: hoveredItem.id } });
+      }
+
+      normalizeAfterRemoval(change);
```
OAGr: I feel like the above functions have become quite long. I'm more uncomfortable with messiness in the actions than I am in other areas, because these can be a pain to debug. 

Perhaps there could be a pattern for mutations to change. I could imagine code like,
```js
changeMutations = [correctWrapping, normalizeAfterRemoval]
applyChanges(value, changeMutations)
```

client/src/pages/RootWorkspacePage/index.tsx
```diff
@@ -75,14 +75,14 @@ const RootWorkspace = ({ workspace }) => {
   return (
     <WorkspaceStyle>
       <Link to={`/workspaces/${workspace.id}`}>
-        <RootBlock block={question} />
+        <RootBlock block={question} availablePointers={[]} />
```
OAGr: could [] be the default value for the prop ``availablePointers`` instead, perhaps?

client/src/utils/slate/closestPointerExportAncestor.ts
```diff
@@ -0,0 +1,15 @@
...
+  let curNode = node;
+  while (
+    !(curNode.type && curNode.type === "pointerExport")
+    &&
+    document.getParent(curNode.key)
+  ) {
+    curNode = document.getParent(curNode.key);
+  }
+
+  if (curNode.type && curNode.type === "pointerExport") {
```
OAGr: I personally have liked using ``_.get`` for doing this, it's optional of course.

client/src/slate-helpers/slate-change-mutations/index.js
```diff
@@ -0,0 +1,17 @@
+import { adjustCursorIfAtEdge } from "./adjustCursorIfAtEdge";
```
stuhlmueller: Does `export { ... } from ...` work?

OAGr:
I'm pretty happy with this at this point. Could imagine some obvious further cleanup, but I suggest just keeping that in mind for future refactors and changes. Happy to merge once @stuhlmueller is satisfied.

client/src/slate-helpers/slate-change-mutations/index.js
zjmiller: Yeah I think that makes more sense. I'll go ahead and make the change.

stuhlmueller:
Happy to merge. I noticed two surprising things while testing this PR, #155 and #156, but we can address them separately. (Not sure when they were introduced.)

Issue #149, opened 2018-08-01
Avoid Auth0 rate limiting errors


stuhlmueller:
I've been encountering this error message (in the console) regularly while editing. I edit, the error sign shows up:

<img width="722" alt="screen shot 2018-08-01 at 4 49 36 pm" src="https://user-images.githubusercontent.com/382515/43554726-edca6350-95aa-11e8-980c-ec837a7808db.png">

Then I check the console and it looks like this:

<img width="946" alt="screen shot 2018-08-01 at 4 50 30 pm" src="https://user-images.githubusercontent.com/382515/43554742-0cec3bb4-95ab-11e8-8b2d-519f215f75aa.png">

stuhlmueller:
When I continue writing, it generally works again, so the issue is either sporadic or has something to do with rate limiting.

stuhlmueller:
I'm surprised we're running into this issue. We ran into rate limiting before, but then we switched to Auth0's paid plan which as far as I remember allows 50 requests / second, which should be plenty.

stuhlmueller:
Closed in #159

PR #157, opened 2018-08-31
Square bracket export

zjmiller:


stuhlmueller:
Nice! This makes creating lists much easier.

One thing I noticed is that - once the pointers are created - the cursor doesn't always end up at the end of the list (where it starts out). 

brackets-cursor

Would this be easy to fix?

(Also, I'm constantly running into #149, the Auth0 rate limiting issue.)

zjmiller:
Trying to "save" the original selection and "reload" it at the end of the exporting process turned out to be really complicated, since the structure of the content is changing throughout the process and Slate's way of encoding cursor position is sensitive to such changes.

So I ended up just modifying the order in which the exporting occurs. Instead of outside-first, inside-last exporting now occurs inside-first, outside-last, and the cursor naturally ends the process at the end of the list.

stuhlmueller:
That's much better, thanks!

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -195,6 +196,26 @@ export class BlockEditorEditingPresentational extends React.Component<
...
+    const didConvertBrackets = slateChangeMutations.scanBlockAndConvertOuterSquareBrackets({
+      change: this.props.value.change(),
+      updateBlock: this.props.updateBlock,
+      exportSelection: this.props.exportSelection,
+      blockId: this.props.block.id,
+    });
+
+    // if something was exported, redo this process
+    if (didConvertBrackets) {
+      setTimeout(() => this.handleSquareBracketExport(), 10);
```
OAGr: I don't believe you need that inner function if ``handleSquareBracketExport`` doesn't take any args. You can just write it as (I think),
```js
 setTimeout(this.handleSquareBracketExport, 10);
```

client/src/modules/blockEditor/actions.ts
```diff
@@ -56,12 +56,15 @@ export const changePointerReference = ({ id, reference }) => {
...
+export const exportSelection = blockId => {
   return async (dispatch, getState) => {
     const { blocks, blockEditor } = await getState();
     const { hoveredItem } = blockEditor;
 
-    const block = blocks.blocks.find(b => b.id === hoveredItem.blockId);
+    const block =
+      blocks.blocks.find(b => b.id === blockId)
+      ||
+      blocks.blocks.find(b => b.id === hoveredItem.blockId);
```
OAGr: I think the top three lines could be cleaned up to something like, 
```js
blocks.blocks.find(b => b.id === blockId || b.id === hoveredItem.blockId)
```

client/src/components/BlockEditor/BlockEditorEditing.tsx
OAGr: Have you thought much about ways to do this without the recursion here? It seems pretty messy to me, though I'd have to spend some time with it to get a better sense.

client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -195,6 +196,26 @@ export class BlockEditorEditingPresentational extends React.Component<
...
 
+  private handleSquareBracketExport() {
+    // check to see whether there are a balanced number of square brackets
+    // if there are, everything within the outermost brackets gets exported
+    const didConvertBrackets = slateChangeMutations.scanBlockAndConvertOuterSquareBrackets({
+      change: this.props.value.change(),
+      updateBlock: this.props.updateBlock,
+      exportSelection: this.props.exportSelection,
+      blockId: this.props.block.id,
+    });
```
OAGr: I'm not a big fan of the pattern of mutations return true or false depending on if they ran or not; it's just not obvious to me that ``scanBlockAndConvertOuterSquareBrackets`` would either run or not run, and would return a boolean if it did or not. I generally expect mutations to be silent.

Not sure what the best way to deal with this is. Perhaps if it returned a value like, {wasMutationPerformed: true} instead.

client/src/slate-helpers/slate-change-mutations/scanBlockAndConvertOuterSquareBrackets.js
```diff
@@ -0,0 +1,118 @@
...
+          startOffset = i;
+        }
+      } else if (chars[i] === "]") {
+        squareBracketBalance--;
+        if (!matchFound) {
+          endKey = textNode.key;
+          endOffset = i;
+          matchFound = true;
+        } else if (squareBracketBalance < 0) {
+          return { balancedBrackets: false };
```
OAGr: I don't like the name ``balancedBrackets``; I would have expected a non-bool value from that. Maybe ``hasBalancedBrackets`` or similar.

client/src/slate-helpers/slate-change-mutations/scanBlockAndConvertOuterSquareBrackets.js
```diff
@@ -0,0 +1,118 @@
...
+
+function findOutermostSquareBrackets(textNodes) {
+  let startKey;
+  let startOffset;
+  let endKey;
+  let endOffset;
+
+  let matchFound = false;
+  let squareBracketBalance = 0;
+
```
OAGr: I imagine there are ways to clean up the for/while loop below, but don't know if it's worth the time. If you can figure out something clever then kudos. One nice thing is that at least this function is very contained.

OAGr:
Finished leaving comments. Overall good I think, I like the separation. Not sure about the loop. There's definitely some pretty big interfaces and messy-ish stuff going around. Kudos for this version.

client/src/slate-helpers/slate-change-mutations/scanBlockAndConvertOuterSquareBrackets.js
zjmiller: I'll just leave it as it is for now, but if you think of a concrete improvement I'd be happy to incorporate it.

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: I'll go with returning `{wasMutationPerformed: true}`. I don't think it's totally out there to return something to indicate whether a mutation occurred. For example, in Sequelize, the `destroy` method returns an integer indicating the number of destroyed rows. That said, I'd be happy if we could think of a more natural way to do this.

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: Yeah I explored a few different approaches. This is definitely the most straightforward and least brittle. The central issue is that, given the way that app is currently structured, exports happen one at a time, and exports are based on a selection. So when you have multiple exports that you want to do, you run into the following difficulty. After the first exports occurs (since they have to happen one at a time), the structure of the document has changed, and all of your previous selection information is obsolete. This makes it really tricky to do anything other than the (kinda awkward) approach taken here: perform one export and then take a look at the new document to get the selection info for the next export.

client/src/components/BlockEditor/BlockEditorEditing.tsx
OAGr: This works. I'll kind of keep it in mind, but don't expect to reach an epiphany in the next few days or so.

OAGr:
Looks good to me

PR #160, opened 2018-09-04
Prevent exports across pointer boundaries

zjmiller:
This change prevents attempts to export across pointer boundaries. This should eliminate the split pointer bug, where sometimes the same pointer would get split up into multiple parts.

### DIFF
--- a/client/src/slate-helpers/slate-change-mutations/insertPointerExport.ts
+++ b/client/src/slate-helpers/slate-change-mutations/insertPointerExport.ts
@@ -1,6 +1,7 @@
 import * as uuidv1 from "uuid/v1";
 import { Change } from "../../components/BlockEditor/types";
 import * as slateChangeMutations from "../../slate-helpers/slate-change-mutations";
+import { isSelectionAcrossPointers } from "../slate-utils/isSelectionAcrossPointers";
 
 const onlyOneNodeThatIsPointerExport = nodes => {
   return (
@@ -35,6 +36,14 @@ export function insertPointerExport(change: Change) {
   // pick up what comes after
   if (isNestedInPointerExport) {
     slateChangeMutations.moveSelectionAwayFromPointerEdge(change);
+  }
+
+  // disallow attempts to export across pointer boundaries
+  const selection = change.value.selection;
+  const document = change.value.document;
+  const isExportingAcrossPointers = isSelectionAcrossPointers(selection, document);
+  if (isExportingAcrossPointers) {
+    return;
   }
 
   // A Slate fragment is a document: value.fragment is the document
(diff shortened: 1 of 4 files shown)


client/src/slate-helpers/slate-change-mutations/insertPointerExport.ts
```diff
@@ -37,6 +38,14 @@ export function insertPointerExport(change: Change) {
     slateChangeMutations.moveSelectionAwayFromPointerEdge(change);
   }
 
+  // disallow attempts to export across pointer boundaries
+  const selection = change.value.selection;
+  const document = change.value.document;
+  const isExportingAcrossPointers = isSelectionAcrossPointers(selection, document);
```
OAGr: what about passing ``value`` in instead, then not having to make those two const declarations above? Are you expecting to use ``isSelectionAcrossPointers`` in areas where value is not available?

client/src/slate-helpers/slate-utils/isSelectionAcrossPointers.js
```diff
@@ -0,0 +1,16 @@
...
+  const focusNode = document.getNode(focusKey);
+
+  const closestPointerExportAncestorOfAnchorNode = closestPointerExportAncestor(anchorNode, document);
+  const closestPointerExportAncestorOfFocusNode = closestPointerExportAncestor(focusNode, document);
+
+  if (closestPointerExportAncestorOfAnchorNode !== closestPointerExportAncestorOfFocusNode) {
+    return true;
+  }
+
+  return false;
```
OAGr: I find the above if / return / return pattern a bit awkward. Would prefer something like,
```js
return (closestPointerExportAncestorOfAnchorNode !== closestPointerExportAncestorOfFocusNode)
```

client/src/slate-helpers/slate-utils/isSelectionAcrossPointers.js
```diff
@@ -0,0 +1,16 @@
+import { closestPointerExportAncestor } from "./closestPointerExportAncestor";
+
+export function isSelectionAcrossPointers(selection, document) {
+  const { anchorKey, focusKey } = selection;
+  const anchorNode = document.getNode(anchorKey);
+  const focusNode = document.getNode(focusKey);
+
+  const closestPointerExportAncestorOfAnchorNode = closestPointerExportAncestor(anchorNode, document);
+  const closestPointerExportAncestorOfFocusNode = closestPointerExportAncestor(focusNode, document);
```
OAGr: From some testing, it seems like this returns undefined if this the anchor/focus are not in an export. If this is the case, then ``closest`` seems a bit misleading. Maybe something like, ``containingExportAncestor``?

OAGr:
I think this is a pretty good solution. Left some comments on syntax. I'm fine with it in general.

OAGr:
Good changes, looks good to me.

PR #161, opened 2018-09-04
Allow multi-paragraph copy-paste

zjmiller:
Previously, when copy-pasting multiple paragraphs, only the first
paragraph would get saved to the server. This was because the app
assumed each Slate document had only one block child. I generalized
so that this assumption is no longer made. Now a Slate document might
have a single block node as a child, or many. I also refactored
minimally by moving the client-side getAllInlinesAsArray into its own
file.

### DIFF
--- a/client/src/lib/slateParser.ts
+++ b/client/src/lib/slateParser.ts
@@ -1,7 +1,7 @@
 import Plain from "slate-plain-serializer";
 
 export function valueToDatabaseJSON(value: any) {
-  return JSON.stringify(value.toJSON().document.nodes[0]);
+  return JSON.stringify(value.toJSON().document.nodes);
 }
 
 export function databaseJSONToValue(databaseJson: any) {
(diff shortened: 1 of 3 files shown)


client/src/lib/slateParser.ts
```diff
@@ -11,7 +11,7 @@ export function databaseJSONToValue(databaseJson: any) {
       document: {
         object: "document",
         data: {},
-        nodes: [databaseJson]
+        nodes: [].concat(databaseJson),
```
OAGr: Why was this change made? What shape is databaseJson?

server/lib/utils/slateUtils.ts
```diff
@@ -1,9 +1,16 @@
 import * as _ from "lodash";
 
-export const getAllInlinesAsArray = node => {
+export const getAllInlinesAsArray = nodeOrNodes => {
```
OAGr: I think some types would really help with things like this. Maybe not now though.

OAGr:
This looks good to me, I'm fine with it being accepted.

client/src/lib/slateParser.ts
OAGr: Oh, good to know, thanks.

PR #162, opened 2018-09-06
Show correct export # in new subquestion form

zjmiller:
Previously, exports in the new subquestion form weren't added to the "available pointers", and hence not properly numbered. This PR addresses this issue.


client/src/slate-helpers/slate-utils/getInlinesFromNodeAndDescendants.ts
```diff
@@ -0,0 +1,13 @@
+export function getExportsFromNodeAndDescendants(node: any, pointerExportsSoFar: any[] = []) {
```
OAGr: This reads like you're mutating an input.

client/src/slate-helpers/slate-utils/getInlinesFromNodeAndDescendants.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (node.object === "inline" && node.type === "pointerExport") {
+    pointerExportsSoFar = pointerExportsSoFar.concat(node);
```
OAGr: you could also do pointerExportsSoFar.push(node)

client/src/slate-helpers/slate-utils/getInlinesFromNodeAndDescendants.ts
zjmiller: Yeah good catch. This went through some iterations and I somehow missed this. I'll clean it up.

OAGr: a specific type declaration for pointerExportsSoFar, and maybe even for node, would be useful.

client/src/components/NewBlockForm.tsx
```diff
@@ -5,12 +5,48 @@ import { valueToDatabaseJSON } from "../lib/slateParser";
...
+    if (this.numOfPointerExports !== pointerExports.length) {
+      this.props.setExportsInNewQuestionForm(pointerExports);
+      this.numOfPointerExports = pointerExports.length;
+
+      // we can return false here because the change to availablePointers
+      // up the hierarchy will trigger a new update
+      return false;
+    }
+
+    return true;
```
OAGr: This seems like a pretty messy solution to me. Generally, data shouldn't be passed up from a child to a parent, like this. It's duplicating state. 

I suggest instead moving the data from the new form into redux, rather than having blockvalue stay in it. This wasn't needed before, but feels like it's needed now.

OAGr: Seperately, I like that you left comments, but would have preferred small functions with nice names :)

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -134,6 +134,13 @@ export class FormPagePresentational extends React.Component<any, any> {
 
   public constructor(props: any) {
     super(props);
+    this.state = {
+      pointersExportedInNewQuestionForm: [],
+    };
+  }
+
+  public setExportsInNewQuestionForm = pointersExportedInNewQuestionForm => {
+    this.setState({ pointersExportedInNewQuestionForm });
```
OAGr: I like the name /size/syntax of this function.

OAGr:
yay, that's so much better. Nice work.

PR #165, opened 2018-09-09
Add client-side timer for workspaces

zjmiller:


zjmiller:
Check out the timer functionality by adding `/timed/10000` after a workspace URL for a 10 second countdown. 

So, for example: `https://mosaic-prod-pr-165.herokuapp.com/workspaces/95d4ac43-f364-4fab-9fd3-c916f92bbfc0/timed/10000`

Change the number at the end of the URL to however many milliseconds you'd like the timer to be. Currently, the only way to reset the timer is to manually clear the item from your `localStorage`. I figure this is fine for now because in experiments we won't want the same person editing the same workspace more than once, but it does make it a bit tricky for testing purposes.

client/src/lib/timer/howMuchTimeLeftOnTimer.js
```diff
@@ -0,0 +1,5 @@
+export default function howMuchTimeLeftOnTimerInMs() {
+  const curTimeInMs = Date.now();
+  const timeLeftInMs = localStorage.getItem('mosaic-timer') - curTimeInMs;
```
OAGr: I highly recommend moving any interaction with localstorage out of libraries. Make it an argument, and make a non-lib function call it with the localStorage reference. 

Also, I don't think it's worth it to have all these tiny timer functions in different files.

client/src/lib/timer/hasTimerStarted.js
```diff
@@ -0,0 +1,11 @@
+export default function hasTimerStarted() {
+  const timerValue = localStorage.getItem('mosaic-timer');
+  if (
+    timerValue !== undefined
+    &&
+    timerValue !== null
+    &&
+    !isNaN(Number(timerValue))
```
OAGr: I suggest checking out the lodash functions for things like this. They have ``isFinite`` and similar which should do what you want.

client/src/lib/timer/howMuchTimeLeftOnTimer.js
OAGr: Also, the string 'mosaic-timer' should definitely be a constant in one place.

OAGr: Would this timer be independent of task? Like, if a user started one task for 5 minutes, then came back for another task for some reason, would it be persisted somehow?

zjmiller: The answer to your last question is yes.

OAGr: In that case I may suggest making the timer task or workspace specific, to be safe. Like {mosaicTimer: {"sdfsdf": 333}}

client/src/pages/EpisodeShowPage/Timer.tsx
```diff
@@ -0,0 +1,45 @@
...
+          m
+          {" "}
+          <span
+            style={{
+              fontSize: "28px",
+            }}
+          >
+            {moment(this.props.timeLeft).format("ss")}
+          </span>
+          s
```
OAGr: If the goal is to have this formatted similar to "30m10s" I believe there is a way to do that with moment's formatting.

OAGr: Separately I've recently found out about Luxon, which may be better than moment for us (It may have trouble with older browsers by do not think that is a big deal)
https://moment.github.io/luxon/

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -134,9 +140,22 @@ export class FormPagePresentational extends React.Component<any, any> {
...
 
   public componentDidMount() {
+    if (this.props.match.params.timeLimit) {
+      if (!hasTimerStarted()) {
+        startTimer(Number(this.props.match.params.timeLimit));
+      }
+
+      setInterval(() => {
+        this.setState({timeLeft: howMuchTimeLeftOnTimer()});
+      }, 250);
```
OAGr: I suggest canceling this setInterval once the component is unmounted.
https://stackoverflow.com/questions/109086/stop-setinterval-call-in-javascript

OAGr: Another example: https://codepen.io/smonn/pen/KzezEw

client/src/lib/timer/howMuchTimeLeftOnTimer.js
zjmiller: Oh yeah, definitely.

zjmiller:
Ok the format is now a bit different. Instead of putting the number of milliseconds in the URL (which seemed a bit error-prone in the sense it would be easy to be off by a 0), you now use a duration string that `moment.duration()` can accept. These include colon-separated hour, minute, second strings, as well as any string conforming to ISO 8601. So, for example, to have timed workspace with five minutes and 30 seconds, end the url with `/timed/0:05:30` or `/timed/0:5:30`.

client/src/pages/EpisodeShowPage/Timer.tsx
zjmiller: I briefly looked into Luxon. I was especially interested in whether it had more options for parsing strings into durations. But it didn't, and for our purposes seems close enough to moment. Advantages concerning time zones and i18n aren't really relevant for durations. Will keep in mind though.

OAGr:
Why is a time of the timer not a query parameter?

client/src/pages/EpisodeShowPage/Timer/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+        {" "}
+        <span
+          style={{
+            fontSize: "28px",
+          }}
+        >
+          {moment(this.props.timeLeft).format("ss")}
+        </span>
+        s
+      </span>
```
OAGr: I suggest writing a function with string interpolation to do the above,  if moment JS can't do it. 

let convert(time) = `${}m ${s}s`...

client/src/pages/EpisodeShowPage/Timer/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  public state = {
+    timeLeft: Infinity,
+  };
+
+  private localStorageKey;
+  private pollingInterval;
+
+  public constructor(props: any) {
+    super(props);
+    this.localStorageKey = `mosiac-timer-${props.workspaceId}`;
```
OAGr: Sorry to say this, but I'm not sure if this is the best technique either. I would expect that this may produce lots of local storage entries that may be difficult to delete and manage.   Right now I believe that these aren't ever removed, which would be a nice addition,  but may not be enough.

I think I suggest something like having a more formal data structure for this;  have the same stuff but nested inside a key, like "mosaic". It would be more code but it would make debugging easier.

OAGr: In general I like to be extra clean & careful when dealing with local storage. It's an external entity  and is relatively important to be understandable for debugging purposes.

PR #166, opened 2018-09-11
Automated task scheduling

zjmiller:
I'm putting this up here just so people can test out the development build. I still need to refactor so please don't review quite yet.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -14,6 +14,8 @@ import thunk from "redux-thunk";
 import "bootstrap/dist/css/bootstrap.min.css";
 
 import { EpisodeShowPage } from "./pages/EpisodeShowPage";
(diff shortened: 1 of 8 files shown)


zjmiller:
I was planning to have a separate PR to isolate the workspace. That should be pretty straightforward (mostly just removing buttons/links).

I'll see if I can rework the code a bit to see if I can make it easier to follow. One thing to note is that the scheduler doesn't check to see whether a workspace has _evidence_ of being worked on (e.g., text in the scratchpad, child questions, etc.), but rather just checks to see if it's scheduled a user to work on it (that could explain your observation in the last bullet point). 

Also, currently the scheduler just stores the schedule information in memory, so a server restart will reset it. I plan on adding persistence soon.

PR #168, opened 2018-09-13
Add isolated workspace

zjmiller:


zjmiller:
So I changed `/workspaces/` to `/isolated-workspaces/` in the route for these instead of adding something to the end, because the timed workspaces are indicated with `/timed/` at the end. If we want to overhaul how we're thinking of these routes and URLs that should be very easy to change.

OAGr:
Why aren't we using query parameters for this?  I'm really hesitant to have what are essentially Boolean arguments as url routes.   That type of thinking could get messy pretty quickly, as evidence in part because of the need to do things like this.

PR #172, opened 2018-09-18
DON'T REVIEW: Demo branch w/ initial style redesign

zjmiller:


zjmiller:
OK so this is connected to the production database, so you don't need to create any trees to get a feel for the style redesign.

zjmiller:
Getting "next workspace" doesn't work for me when connected to the production database. My guess (especially given the 503 error) is that this is a scalability issue related to how many workspaces exist. My automated scheduling algorithm assumes all workspaces are eligible, and in addition I'm sure isn't remotely close to the most efficient implementation.

I'll be looking into this more, but the automated scheduling algorithm roughly set up as it is might be sufficient for near-term multi-user experiments.

PR #196, opened 2018-10-04
Automated scheduling v1

zjmiller:
Also includes tests


server/tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "noImplicitAny": false,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
-    "noUnusedLocals": false
+    "noUnusedLocals": false,
+    "downlevelIteration": true
```
zjmiller: This needs to be true so that TypeScript properly transpiles `for... of...` loops that iterate over Maps.

server/test/lib/scheduler/Assignment.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it("can retrieve workspace id", () => {
+    expect(assignment.getWorkspaceId()).to.equal(WORKSPACE_ID);
+  });
+
+  it("records initial timestamp", () => {
+    expect(assignment.getStartedAtTimestamp()).to.be.a("number");
+    const curTimestamp = Date.now();
+    expect(curTimestamp - assignment.getStartedAtTimestamp() > 1000);
```
OAGr: Do you need to have ``.to.equal`` here? Also not sure if this tests what it says.

server/test/lib/scheduler/Assignment.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it("records correct timestamp", () => {
+    const curTimestamp = Date.now();
+    expect(curTimestamp - assignment.getStartedAtTimestamp() < 1000).to.equal(true);
```
OAGr: Is this testing that these two are "close enough" to each other, where "close enough" is within one second? If so make that more explicit and add comments. Seems a bit sloppy.

server/test/lib/scheduler/Assignment.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  beforeEach(() => {
+    assignment = new Assignment({
+      userId: USER_ID,
+      workspaceId: WORKSPACE_ID,
+    });
+  });
+
+  it("has tests that work", () => {
+    expect(true).to.equal(true);
+  });
```
OAGr: How strongly do you feel about having this in every test file?

server/test/lib/scheduler/Schedule.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+  });
+
+  it("has tests that work", () => {
+    expect(true).to.equal(true);
+  });
+
+  // tests private method, maybe remove
+  describe("doesUserHaveASchedule", () => {
+    it("works when user doesn't have a schedule", () => {
```
OAGr: I suggest using ``contexts`` as a second layer to make this a bit more specific, a la Ruby.

http://www.albertgao.xyz/2016/12/09/testing-via-mocha-and-chai-in-es5-and-es6/

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,143 @@
...
+      const curTimestamp = userSchedule.getTimestampWorkspaceLastWorkedOn(workspaceId);
+      if (mostRecentTimestamp < curTimestamp) {
+        mostRecentTimestamp = curTimestamp;
+      }
+    }
+
+    return mostRecentTimestamp;
+  }
+
+  public getWhichOfTheseWorkspacesWorkedOnLeastRecently(workspaceIds) {
```
OAGr: maybe ``getMostRecentlyActiveWorkspace`` or similar?

server/test/lib/scheduler/Schedule.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+      const result = schedule.getWhichOfTheseWorkspacesWorkedOnLeastRecently([WORKSPACE_ID_1, WORKSPACE_ID_2]);
+      expect(result).to.equal(WORKSPACE_ID_2);
+    });
+  });
+
+  describe("getTreesWorkedOnLeastRecently", () => {
+    it("works in straightforward case", done => {
+      schedule.assignWorkspaceToUser(USER_ID_1, "1-1");
+      fakeClock.tick(ONE_MINUTE);
+      schedule.assignWorkspaceToUser(USER_ID_2, "2");
```
OAGr: I suggest pulling out the Time.now() functionality in assignWorkspaceToUser, so it is an optional arg and can be passed in manually. That way this testing could be cleaner.

server/test/lib/scheduler/Scheduler.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+          ]);
+          done();
+        }).catch(e => {
+          done(e);
+        });
+      });
+    });
+  });
+
+});
```
OAGr: You can definitely DRY up a bunch of these tests. No need to stay so close to native chai syntax.

server/lib/scheduler/RootParentFinder.ts
```diff
@@ -0,0 +1,47 @@
+import * as models from "../models";
```
OAGr: Should this be in the scheduler? It feels very general purpose to me, I'm surprised it wasn't needed earlier. Seems like it should be a method on ``workspace`` (``#rootParent``)

server/lib/scheduler/RootParentFinder.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const workspace = await models.Workspace.findById(workspaceId);
+
+    if (!workspace.parentId) {
+      this.rootParentCache[workspaceId] = workspace;
+      return workspace.id;
+    } else {
+      const rootParent = await this.getRootParentOfWorkspace(workspace.parentId);
+      this.rootParentCache[workspaceId] = rootParent;
+      return rootParent.id;
```
OAGr: This functionality seems very generic, like something I would expect in the workspace model file.

server/lib/scheduler/RootParentFinder.ts
```diff
@@ -0,0 +1,47 @@
+import * as models from "../models";
+
+class RootParentFinder {
+  private static RootParentCache = {};
+
+  public static clearRootParentCache() {
+    this.rootParentCache = {};
+  }
+
+  public static async getRootParentOfWorkspace(workspaceId, workspace) {
```
OAGr: This doesn't feel right to me. I think this file should be called ``RootParentCache``, but even then, maybe could be replaced by a redis cache with a more general purpose solution.

server/lib/scheduler/RootParentFinder.ts
zjmiller: As you might have guessed, this is here just because caching these "root parent" results is required for the scheduler to work. Without the caching here, performance is prohibitively bad.

zjmiller: Not sure if it makes sense to put caching on the model.

server/lib/scheduler/RootParentFinder.ts
zjmiller: Name change sounds good.

server/test/lib/scheduler/Assignment.spec.ts
zjmiller: I can take them all out. They were useful when setting everything up but I agree that we don't need them now.

server/test/lib/scheduler/Assignment.spec.ts
zjmiller: This works as it is (don't need `.to.equal`). But I'll try to make it less awkward.

server/lib/scheduler/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Scheduler } from "./Scheduler";
+
+const ONE_MINUTE = 60 * 1000;
+
+//  By separating out fetchAllWorkspaces and rootParentCache, we 100% isolate
+//  the rest of the scheduling code (Scheduler, Schedule, UserSchedule, and
+//  Assignment classes) from Sequelize & Postgres
+
+const scheduler = new Scheduler({
+  fetchAllWorkspaces: () => Workspace.findAll(),
```
OAGr: I think this is the same as 
```
  fetchAllWorkspaces: Workspace.findAll,
```

server/lib/scheduler/UserSchedule.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public hasUserWorkedOnWorkspace(workspaceId) {
+    for (const assignment of this.userSchedule) {
+      if (assignment.getWorkspaceId() === workspaceId) {
+        return true;
+      }
+    }
+
+    return false;
```
OAGr: I think this can be cleaned up with lodash. Something like,
```
_.some(this.userSchedule, {workspaceId})
```
(I'm not sure if _.some takes a function, but you get the idea. That's also one disadvantage of hiding the workspaceId behind a method.)

OAGr: Actually, that read is very cheap, so if you had to first map, then _.some, that would probably be fine.

server/lib/scheduler/UserSchedule.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    const howLongAgoUserStartedWorkingOnIt = Date.now() - lastWorkedOnAssignment.getStartedAtTimestamp();
+    const didUserStartWorkingOnItWithinTimeLimit = howLongAgoUserStartedWorkingOnIt < this.timeLimit;
+
+    if (didUserStartWorkingOnItWithinTimeLimit) {
+      return true;
+    }
+
+    return false;
```
OAGr: I'd split two private functions from this, ```lastWorkedOnWorkspaceId()``` and ```isActiveInLastWorkspace```. Then this method would be something like ```lastWorkedOnWorkspaceId() === workspaceId && isActiveInLastWorkspace()```

server/lib/scheduler/UserSchedule.ts
```diff
@@ -0,0 +1,75 @@
...
+    let mostRecentTimestamp = -Infinity;
+
+    for (const assignment of this.userSchedule) {
+      if (assignment.getWorkspaceId() === workspaceId) {
+        const curTimestamp = assignment.getStartedAtTimestamp();
+        if (mostRecentTimestamp < curTimestamp) {
+          mostRecentTimestamp = curTimestamp;
+        }
+      }
+    }
```
OAGr: Again, I prefer the style where you use lodash for this with filters and the like. I bet you could find something pretty short.

OAGr: Btw, Ramda seems even better than Lodash, but I don't know it that well. If you prefer that I would recommend it.

server/lib/scheduler/Assignment.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  public getStartedAtTimestamp() {
+    return this.startAtTimestamp;
+  }
+
+  public getWorkspaceId() {
+    return this.workspaceId;
+  }
+}
```
OAGr: Why did you make this its' own class? If you leave it as an object it actually may be a bit more flexible for querying and so on.

OAGr: If you did so for the Date.now() in the constructor, you could still have a contructor-like function.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+      return [idOfWorkspaceWorkedOnLeastRecently];
+    }
+
+    return finalWorkspaces.map(w => w.id);
+  }
+
+  private pickWorkspaceIdAtRandom(workspaceIds) {
+    const randomIndex = Math.floor(Math.random() * (workspaceIds.length));
+    return workspaceIds[randomIndex];
+  }
```
OAGr: This screams "utility method" to me; you could rename it ```pickRandomItem``` and make it generic for all arrays, and place it above. In part, to make it apparent that what it's doing is generic.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  public async findNextWorkspace(userId) {
+    // clear cache so we don't use old eligibility info
+    this.rootParentCache.clearRootParentCache();
+    const idsOfWorkspacesThatCouldBeNext = await this.getIdsOfWorkspacesThatCouldBeNext(userId);
+    const assignedWorkspaceId = this.pickWorkspaceIdAtRandom(idsOfWorkspacesThatCouldBeNext);
+    await this.schedule.assignWorkspaceToUser(userId, assignedWorkspaceId);
+  }
+
+  private async getIdsOfWorkspacesThatCouldBeNext(userId) {
```
OAGr: I don't think this needs the param userId, it doesn't seem to use it.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.rootParentCache = rootParentCache;
+    this.schedule = schedule;
+  }
+
+  public async getIdOfCurrentWorkspace(userId) {
+    const assignment = this.schedule.getMostRecentAssignmentForUser(userId);
+    return assignment.getWorkspaceId();
+  }
+
+  public async findNextWorkspace(userId) {
```
OAGr: I suggest renaming this to something like ``getNextBestWorkspaceForAssignment`` or similar. Then I'd remove the userId part of this and line 24, and have them either in a different function or outside of this class.

server/lib/scheduler/Scheduler.ts
OAGr: Also, why have this function only return the Ids? If it returns the workspaces, it's easy to filter on them, and the name would be shorter here (this seems done in a few places)

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private async filterByEligibility(workspaces) {
+    // use for... of instead of asyncro's filter
+    // because isWorkspaceEligible will use rootParentCache
+    // and the parallel nature of filter doesn't work with
+    // rootParentCache's caching
+    let allEligibleWorkspaces = [];
+    for (const workspace of workspaces) {
+      const isMarkedEligible = await this.isWorkspaceEligible(workspace);
```
OAGr: I don't think this needs to be an await

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+      const isCurrentlyBeingWorkedOn = await this.schedule.isWorkspaceCurrentlyBeingWorkedOn(workspace.id);
```
OAGr: Also don't this this needs an await? If that's the case, maybe this section could be done using lodash or something simpler.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  /*
+    NOTE: this only considers the trees that at least one workspace in the
+    workspaces argument belong to
+  */
+  private async filterByWhetherInTreeWorkedOnLeastRecently(workspaces) {
+    let workspacesInTreeWorkedOnLeastRecently = [];
+
+    for (const workspace of workspaces) {
+      const shouldInclude = await this.schedule.isInTreeWorkedOnLeastRecently(workspaces.map(w => w.id), workspace.id);
```
OAGr: If this is needed, it would make sense to me for ``this.schedule`` to expose the full list of workedOnRecently workspaces, and then this could be a simple comparison filter.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  private async filterByWhetherNotYetWorkedOn(workspaces) {
+    return await filter(
+      workspaces,
+      async w => !await this.schedule.hasWorkspaceBeenWorkedOnYet(w.id)
+    );
+  }
+
+  private filterByWhetherHasRemainingBudget(workspaces) {
+    return workspaces.filter(w => this.hasRemainingBudgetForChildren(w));
```
OAGr: should be fine with 
```
    return workspaces.filter(this.hasRemainingBudgetForChildren);
```

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+      const shouldInclude = await this.schedule.isInTreeWorkedOnLeastRecently(workspaces.map(w => w.id), workspace.id);
+      if (shouldInclude) {
+        workspacesInTreeWorkedOnLeastRecently.push(workspace);
+      }
+    }
+
+    return workspacesInTreeWorkedOnLeastRecently;
+  }
+
+  private async filterByWhetherNotYetWorkedOn(workspaces) {
```
OAGr: After some thinking, I think one decent way to do this filtering would be with Ramda.js. You should be able to compose multiple filtering methods. 

One thing is that instead of having these filter methods that accept arrays of workspaces, I suggest having filter methods that work on individual workspaces. ``fooWorkspaces.filter(WhetherNotYetWorkedOn)`` and similar seem good to me, and with Ramda would be composable (so you could add multiple filters). Let me know if you would like help with that.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+  public async findNextWorkspace(userId) {
+    // clear cache so we don't use old eligibility info
+    this.rootParentCache.clearRootParentCache();
+    const idsOfWorkspacesThatCouldBeNext = await this.getIdsOfWorkspacesThatCouldBeNext(userId);
+    const assignedWorkspaceId = this.pickWorkspaceIdAtRandom(idsOfWorkspacesThatCouldBeNext);
+    await this.schedule.assignWorkspaceToUser(userId, assignedWorkspaceId);
+  }
+
+  private async getIdsOfWorkspacesThatCouldBeNext(userId) {
+    const allWorkspaces = await this.fetchAllWorkspaces();
```
stuhlmueller: Fetching all workspaces and then filtering seems quite inefficient. Any reason not to first find the tree that the user worked on least recently, then only request the workspaces in that tree (and, once time budgets are in, only the ones with remaining budget)?

I guess you'd need to switch to the tree worked on second-least-recently if there are no eligible workspaces in the least-recent one, but you have that issue with the current implementation as well.

server/lib/scheduler/Scheduler.ts
OAGr: Also, I don't like the phrasing "ThatCouldBeNext". Maybe "findActionableWorkspaces" or some other term? ("Valid"?)

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // if every workspace in that tree has 0 remaining budget
+    // then instead look for the workspace worked on least recently
+    if (finalWorkspaces.length === 0) {
+      const idOfWorkspaceWorkedOnLeastRecently = await this.schedule.getLeastRecentlyActiveWorkspace(workspacesInTreeWorkedOnLeastRecently.map(w => w.id));
+      return [idOfWorkspaceWorkedOnLeastRecently];
+    }
+
+    return finalWorkspaces.map(w => w.id);
```
OAGr: This function is really the meat of this class. I think you could make it nicer, for instance, replace ``finalWorkspaces.length === 0`` with ``hasValidCandidates()``

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+    return this.schedule.has(userId);
+  }
+
+  private createUserSchedule(userId) {
+    this.schedule.set(userId, new UserSchedule({ timeLimit: this.timeLimit, userId }));
+  }
+
+  private getUserSchedule(userId) {
+    return this.schedule.get(userId);
+  }
```
OAGr: I like the above small and simple functions, kudos.

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+    this.timeLimit = timeLimit;
+  }
+
+  private doesUserHaveASchedule(userId) {
+    return this.schedule.has(userId);
+  }
+
+  private createUserSchedule(userId) {
+    this.schedule.set(userId, new UserSchedule({ timeLimit: this.timeLimit, userId }));
+  }
```
OAGr: Would it be ok for this to be called multiple times with one userId? If not, I suggest adding a guard, then call this from the ``assignWorkspaceToUser`` fn. Maybe rename to `createUserScheduleIfNotCreated`

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /*
+    NOTE: this only considers the trees that at least one workspace in the
+    argument belong to
+  */
+  public async isInTreeWorkedOnLeastRecently(workspaceIds, workspaceId) {
+    const treesWorksOnLeastRecently = await this.getTreesWorkedOnLeastRecently(workspaceIds);
+    const rootParentId = await this.rootParentCache.getRootParentIdOfWorkspace(workspaceId);
+    return Boolean(treesWorksOnLeastRecently.find(rootWorkspaceId => rootWorkspaceId === rootParentId));
```
OAGr: Haven't seen Boolean used like this much. I'm used to using ```!!foo```, but I guess this should work the same, and is perhaps more straightforward.
That said, with lodash, you should have a fn to do just this.

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+      if (!idOfWorkspaceWorkedOnLeastRecently || lastWorkedOnTimestamp < whenThisWorkspaceWasWorkedOn) {
+        idOfWorkspaceWorkedOnLeastRecently = workspaceId;
+        whenThisWorkspaceWasWorkedOn = lastWorkedOnTimestamp;
+      }
+    });
+
+    return idOfWorkspaceWorkedOnLeastRecently;
+  }
+
+}
```
OAGr: I think the functions above could be cleaned up with Lodash or Rambdajs.

server/lib/schema/index.ts
```diff
@@ -306,7 +307,21 @@ const schema = new GraphQLSchema({
...
+        resolve: async (_, args, context) => {
+          const user = await userFromAuthToken(context.authorization);
+          if (user == null) {
+            throw new Error(
+              "No user found when attempting get next workspace."
+            );
+          }
+          await scheduler.findNextWorkspace(user.user_id);
+          const workspaceId = await scheduler.getIdOfCurrentWorkspace(user.user_id);
+          return { id: workspaceId };
```
OAGr: I like how you made this interface so minimal.

OAGr:
I realize that there is time pressure to merge this. If so, don't worry about most of my comments immediately; but soon after, I do suggest spending some time learning Ramda and then refactoring it or doing that for future modifications.

server/lib/scheduler/Assignment.ts
zjmiller: It just made sense conceptually to me to have an assignment as a stand-alone entity. You could imagine a system in the future that passed assignments around. This is also why I gave them userId's even though this isn't needed the way things are currently set up.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,110 @@
+import { filter } from "asyncro";
+
+class Scheduler {
+  private fetchAllWorkspaces;
```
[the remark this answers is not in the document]
zjmiller: Maybe hold off on this given that server-side type-checking hadn't been occurring? I just fixed this issue in https://github.com/oughtinc/mosaic/pull/199, and unsurprisingly there are a lot of type-related issues I'll need to address.

server/test/lib/scheduler/Scheduler.spec.ts
zjmiller: I think they've been DRYed up some since you posted this. Let me know if there are any specific additional improvements you have in mind.

server/lib/scheduler/UserSchedule.ts
zjmiller: Addressed in 99a270c

server/lib/scheduler/Scheduler.ts
zjmiller: It needs `userId` when it does the assignment.

zjmiller: `findNextWorkspace` is just a mutation. It doesn't return anything. So the name doesn't seem apt, starting with "find".

zjmiller: I'll change it to `assignUserToNextWorkspace`

zjmiller: actually maybe just `assignNextWorkspace`, either way it's now clear it's doing something

server/lib/scheduler/Scheduler.ts
zjmiller: Fair enough, the preview branch version has some async logic in `isWorkspaceEligible()`. This will be coming in the next PR.

server/lib/scheduler/Scheduler.ts
zjmiller: Addressed here: https://github.com/oughtinc/mosaic/pull/196/commits/a8f7044f0a2107ea1db785a72e503a07b3e690c2.  Instead of asking for `fetchAllWorkspaces`, Scheduler now asks for `fetchAllRootWorkspaces`, and `fetchAllWorkspacesInTree`, and avoids performing a query wrt any workspaces not in the least recently worked on trees.

server/lib/scheduler/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### Scheduler
+
+The scheduler is the only exposed API. It only has two public methods: `getIdOfCurrentWorkspace` and `assignNextWorkspace`. The scheduler looks at all of the scheduling information (via the schedule) and decides which workspace to assign to a user next.
+
+## Implementation
+
+There is quite a bit of reasoning about trees. Trees are identified with their root workspaces.
+
+One gotcha to watch out for is that it doesn't appear repeated Sequelize queries preserve object identity. For example doing the following twice `await models.Workspace.findById(workspaceid)` yields distinct objects. This can cause really difficult to detect bugs when using `===` to determine object identity. Right now things are OK because the every time the scheduler finds a next workspace for a user, it clears the cache and fetches Sequelize objects once, so `===` is OK.
```
OAGr: Is this referencing the fact that you test equality for workspaces, instead of for their corresponding IDs? In general I'm pretty paranoid about testing for equality on objects for reasons like this, generally recommend doing it for IDs instead.

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    this.schedule.set(userId, new UserSchedule({ timeLimit: this.timeLimit, userId }));
+  }
+
+  private getUserSchedule(userId) {
+    return this.schedule.get(userId);
+  }
+
+  public async assignWorkspaceToUser(userId, workspace, startAtTimestamp = Date.now()) {
+    if (!this.doesUserHaveASchedule(userId)) {
```
OAGr: This line should not be necessary :)

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+    const userSchedule = this.getUserSchedule(userId);
+    return userSchedule.getMostRecentAssignment();
+  }
+
+  /*
+    NOTE: this only considers the trees associated with the rootWorkspaces
+    argument
+  */
+  public async isInTreeWorkedOnLeastRecently(rootWorkspaces, workspace) {
+    const treesWorksOnLeastRecently = this.getTreesWorkedOnLeastRecently(rootWorkspaces);
```
OAGr: do you mean *WorkedOnLeastRecently?

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
+import { map } from "asyncro";
+import * as _ from "lodash";
+import { UserSchedule } from "./UserSchedule";
+
+class Schedule {
+  // the following object maps root workspace ids to timestamps
+  // I didn't use a Map (and map a workspace object to a timestamp)
+  // because Sequelize doesn't preserve object identity across queries
+  private cacheForWhenTreeLastWorkedOn = {};
```
OAGr: Does this need to have "cache" in its' name? Can it just be something like "lastWorkedOnTime" or similar, with a comment about workspaceId?

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+    return !!(treesWorksOnLeastRecently.find(rootWorkspace => rootWorkspace === rootParent));
+  }
+
+  /*
+    NOTE: this only considers the trees associated with the rootWorkspaces
+    argument
+  */
+  public getTreesWorkedOnLeastRecently(rootWorkspaces) {
+    const treesNotYetWorkedOn = rootWorkspaces.filter(
+      rootWorkspace => this.cacheForWhenTreeLastWorkedOn[rootWorkspace.id] === undefined
```
OAGr: In cases like this, I prefer if the filter uses a var name like ``r`` instead of ``rootWorspace``. We can tell that it is a rootWorkspace from reading 9 characters back.

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    return leastRecentlyWorkedOnTrees;
+  }
+
+  public isWorkspaceCurrentlyBeingWorkedOn(workspace) {
+    for (const [userId, userSchedule] of this.schedule) {
+      if (userSchedule.isUserCurrentlyWorkingOnWorkspace(workspace)) {
+        return true;
+      }
+    }
```
OAGr: For the refactor, this really, really should be changed to a _.some or similar.

server/lib/scheduler/Schedule.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  public getTimestampWorkspaceLastWorkedOn(workspace) {
+    let mostRecentTimestamp = -Infinity;
+
+    for (const [userId, userSchedule] of this.schedule) {
+      const curTimestamp = userSchedule.getTimestampWorkspaceLastWorkedOn(workspace);
+      if (mostRecentTimestamp < curTimestamp) {
+        mostRecentTimestamp = curTimestamp;
+      }
+    }
```
OAGr: the above 8 lines are so should really be changed into a _.min function or similar.

server/lib/scheduler/Scheduler.ts
```diff
@@ -0,0 +1,99 @@
...
+    // then instead look for workspaces with remaining budgets
+    if (finalWorkspaces.length === 0) {
+      finalWorkspaces = await this.filterByWhetherHasRemainingBudget(eligibleWorkspaces);
+    }
+
+    // if every workspace in that tree has 0 remaining budget
+    // then instead look for the workspace worked on least recently
+    if (finalWorkspaces.length === 0) {
+      const workspaceWorkedOnLeastRecently = await this.schedule.getLeastRecentlyActiveWorkspace(eligibleWorkspaces);
+      return [workspaceWorkedOnLeastRecently];
```
OAGr: I recommend keeping to the pattern here, like
```finalWorkspaces = [workspaceWorkedOnLeastRecently]```

server/lib/scheduler/Schedule.ts
zjmiller: K I'll just do it now.

server/lib/scheduler/Schedule.ts
zjmiller: I'll do this now too.

PR #202, opened 2018-10-11
Allow admins to toggle workspace isPublic

zjmiller:


OAGr:
Why toggle instead of update? In general I prefer updating, it makes calls idempotent. (More Rest-full, though I realize this isn't exactly REST)

zjmiller:
This will instead be in a PR working towards getting v0.3 merged into master.

Issue #186, opened 2018-10-01
Make the child budget info less confusing


stuhlmueller:
The display of "used up" time is not really helpful at the moment. If the child passes on its budget to a grandchild, it's displayed as used up, even if the grandchild didn't do anything with it. Not sure what the right answer here is. Maybe we should just not display how much has been used up.

Issue #70, opened 2018-04-17
Refetch pointer values when navigating to parent or child


stuhlmueller:
When navigating to parent, new pointers in responses from children are often n/a.

When navigating to child, new pointers in questions are sometimes n/a.

PR #205, opened 2018-10-16
Preview branch v0.3

zjmiller:


stuhlmueller:
Every now and then when I retrieve the next workspace I get this error:

screen shot 2018-10-17 at 11 37 26 am

The relevant code in index.js:

```js
  public render() {
    const workspace = this.props.workspace.workspace;

    const importedPointers = workspace.connectedPointers;

    const allReadOnlyBlocks = new WorkspaceWithRelations(
      workspace
    ).allReadOnlyBlocks();
    const readOnlyExportedPointers = _.flatten(
      allReadOnlyBlocks.map(b => findPointers(b.value))  // *******
    );
    const availablePointers = _.uniqBy(
      [
        ...this.props.exportingPointers,
        ...importedPointers,
        ...readOnlyExportedPointers
      ],
      p => p.data.pointerId
    );
```

zjmiller:
Thanks, looking into it.

stuhlmueller:
Other feedback:

- Budget remaining should probably be displayed in some more readable form (e.g. 1h30m), and sub-question budgets should accept such readable inputs.

- It would be nice if we could avoid having both the current timer and the total budget count down every second. Here are two possible ways to address this:

  - If the budget remaining is >10m (say), round the display to the nearest minute.
  - Show the budget as a sum, e.g. 30s + 1h50m10s, and only count down the first part. (Would have to be careful with the corner case where the total budget is <30s.) I think this is my preferred option so far.

- For debugging purposes, we could show staleness in the tree view. E.g. by adding a dot next to workspaces that need work.

- If a workspace times out, showing "needs more work" and "done for now" is a bit weird, since the user has lost access to the information they'd need to decide if the workspace needs more work. In that case, maybe there should just be one button "get next workspace" that is equivalent to the "needs more work" button?

zjmiller:
With respect to the last point, trees are represented as root workspaces. So `getTreesWorkedOnLeastRecentlyByUser` just loads every (eligible) root workspace. Then, only after determining which trees the user worked on least recently, are additional non-root workspaces loaded.

stuhlmueller:
Which non-root workspaces are loaded if a user hasn't worked on any trees yet?

zjmiller:
Then all of them, yeah, since all trees are equal wrt when the user last worked on them..

stuhlmueller:
> If multiple trees are equal wrt having been worked on, I could just pick one at random before loading workspaces?

Sounds good!

stuhlmueller:
More notes:

- Another idea for how to adjust the workspace timer: Just say "Time left: yours / this workspace", keeping the information for both timers the same as it is now.

- The display for child budgets should be readable (use hours/minutes/seconds) just like the workspace timer.

- When allocating child budgets, we're still subtracting 1 (which made sense when budgets were about the number of workspaces, but doesn't for time).

- Don't schedule workspaces with less than 10 seconds remaining? It's pretty difficult to do any work in that case.

stuhlmueller:
- Rename "edit allocation" to "edit time budget", or perhaps just "edit time"

- When editing allocation, say "seconds" (or ideally accept the same format as elsewhere)

- Edit the allocation should mark the corresponding child as stale

- When budget is 0 for a subquestino, it currently just says "left" (without number)

- The child budget info should show the total, not the time left (see #186 for justification)

- Under "New Question", the gray default text for budget could give an example of the format (e.g. 5m10s)

- When displaying durations, don't show leading zeros

- I clicked "done for now" in a workspace and it didn't mark the corresponding [workspace](https://mosaic-prod-pr-205.herokuapp.com/workspaces/90f97801-fea9-472b-bc4d-c77b3872cf62) as stale. (I checked staleness in the tree view, which is very helpful!)

client/src/components/NewBlockForm.tsx
```diff
@@ -129,7 +128,7 @@ export class NewBlockForm extends React.Component<any, any> {
   private onSubmit = () => {
     this.props.onMutate({
       question: valueToDatabaseJSON(this.state.blockValue),
-      totalBudget: this.state.totalBudget
+      totalBudget: /^\d+$/.exec(this.state.totalBudget) ? this.state.totalBudget : (parse(this.state.totalBudget) / 1000),
```
zjmiller: Factor out regex into separate function.

client/src/pages/EpisodeShowPage/ChildrenSidebar.tsx
```diff
@@ -50,6 +51,30 @@ export class Child extends React.Component<any, any> {
...
+    }
+
+    if (duration.minutes > 0) {
+      durationString += `${Duration.fromObject({minutes: duration.minutes}).toFormat("m")}m `;
+    }
+
+    if (duration.seconds > 0) {
+      durationString += `${Duration.fromObject({seconds: duration.seconds}).toFormat("s")}s`;
+    }
+
```
zjmiller: Create a separate component for this budget "badge".

client/src/pages/RootWorkspacePage/RootWorkspace.tsx
```diff
@@ -24,51 +33,191 @@ const ScratchpadContainer = styled.div`
...
+          </div>
+        }
+        <Link to={`/workspaces/${workspace.id}`}>
+          <RootBlock
+            availablePointers={workspace.connectedPointers}
+            block={question}
+          />
+        </Link>
+
+        {" "}
```
zjmiller: Make separate components for these checkboxes.

client/src/pages/RootWorkspacePage/RootWorkspace.tsx
```diff
@@ -24,51 +33,191 @@ const ScratchpadContainer = styled.div`
...
+      refetchQueries: ["RootWorkspacesQuery"]
+    }
+  }),
+  graphql(UPDATE_WORKSPACE_IS_PUBLIC, {
+    name: "updateWorkspaceIsPublic",
+    options: {
+      refetchQueries: ["RootWorkspacesQuery"]
+    }
+  })
+)(RootWorkspacePresentational);
```
zjmiller: Think of how to optimistically update checkbox state to improve responsiveness.

server/lib/scheduler/Scheduler.ts
```diff
@@ -83,17 +157,8 @@ class Scheduler {
...
-
-  private isWorkspaceEligible(w) {
-    // will add conditions to this later
-    return true;
-  }
-
-  private hasRemainingBudgetForChildren(workspace) {
-    return workspace.totalBudget > workspace.allocatedBudget;
+  private hasMinRemaining(workspace) {
+    return (workspace.totalBudget - workspace.allocatedBudget) >= 10;
```
zjmiller: Make a constant for this "minimum budget" number (currently at 10).

stuhlmueller:
I just ran into the following issue with a pointer showing up as N/A in the question:

pointer-issue

Note that this is different from #70 in that it persists even after refresh.

server/lib/schema/index.ts
```diff
@@ -163,7 +163,41 @@ const schema = new GraphQLSchema({
       workspace: {
         type: workspaceType,
         args: { id: { type: GraphQLString } },
-        resolve: resolver(models.Workspace)
+        resolve: resolver(models.Workspace, {
+          after: async (result, args, ctx) => {
```
stuhlmueller: When exactly will this hook run?

zjmiller: Whenever the workspace is fetched. Maybe it makes more sense to validate it when it's created. The main problem is I don't really see how this bug arose in the first place, i.e., how a workspace could have been created without having the associated three blocks (question, answer, scratchpad). So here I opted for the solution that's guaranteed to eliminate the bug. You'll never be able to fetch a workspace that's missing one of these required blocks.

stuhlmueller: Would be good to understand how the situation could arise in the first place. (What if the reason it arose can also cause other errors?)

zjmiller: Yeah definitely. If you look the Workspace model's `afterCreate` hook, you'll see the creation of the associated blocks is very straightforward. Also, we've only seen this error once out of 1000+ workspace creations. So my best guess is that understanding how this happened will be very difficult. Having more robust logging on the backend would help w/ future debugging efforts if someone like this happens again.

PR #213, opened 2018-10-22
Admin controls

zjmiller:


server/lib/schema/index.ts
```diff
@@ -289,6 +290,46 @@ const schema = new GraphQLSchema({
...
+        resolve: async (_, { isEligible, workspaceId }, context) => {
+          const user = await userFromAuthToken(context.authorization);
+          if (user == null) {
+            throw new Error(
+              "No user found when attempting to update workspace eligibility."
+            );
+          }
+          const workspace = await models.Workspace.findById(workspaceId);
+          await workspace.update({ isEligibleForAssignment: isEligible });
+          return { id: workspaceId };
```
OAGr: why doesn't this return the full workspace, like the other update has?

server/lib/schema/index.ts
```diff
@@ -289,6 +290,46 @@ const schema = new GraphQLSchema({
           });
         }
       },
+      updateWorkspaceIsPublic: {
```
OAGr: Are you sure about not just having a ``updateWorkspace`` mutation, that accepts multiple optional arguments? I generally recommend staying closer to REST where applicable.

server/lib/schema/index.ts
```diff
@@ -289,6 +290,46 @@ const schema = new GraphQLSchema({
...
+        type: workspaceType,
+        args: {
+          isPublic: { type: GraphQLBoolean },
+          workspaceId: { type: GraphQLString },
+        },
+        resolve: async (_, { isPublic, workspaceId }, context) => {
+          const user = await userFromAuthToken(context.authorization);
+          if (user == null) {
+            throw new Error(
+              "No user found when attempting to toggle workspace visiblity."
```
OAGr: I don't recommend task-specific error names like this. The mutation/task should be obvious in the logs from other info.  I recommend instead making ``checkUser`` or something as one standard method to do authorization.

client/src/pages/RootWorkspacePage/AdminControls/UpdateIsEligibleCheckbox.tsx
```diff
@@ -0,0 +1,90 @@
...
+  private didIsEligibleChange = (prevProps: any, curProps: any) => {
+    const prevWorkspace = prevProps.workspace;
+    const curWorkspace = curProps.workspace;
+    return prevWorkspace.isEligibleForAssignment !== curWorkspace.isEligibleForAssignment;
+  }
+
+  private handleOnIsEligibleCheckboxChange = () => {
+    if (this.state.isEligibleCheckboxStatusPending) {
+      return;
+    }
```
OAGr: Given that this function doesn't return anything, I think you could replace the above 3 lines, and just do the following in an if condition for ``!this.state.isEligibleCheckboxStatusPending``

client/src/pages/RootWorkspacePage/AdminControls/UpdateIsEligibleCheckbox.tsx
```diff
@@ -0,0 +1,90 @@
...
+      // it is only here so the user has a moment to see the "Updating"
+      // message if the server responds very quickly
+      setTimeout(() => {
+        this.props.updateWorkspaceIsEligible({
+          variables: {
+            isEligible: !this.props.workspace.isEligibleForAssignment,
+            workspaceId: this.props.workspace.id,
+          }
+        });
+      }, 200);
```
OAGr: This seems pretty messy to me. I suggest trying to come up with some other way to do so if this file is going to be important in the future; if it's a rush and/or a small/hacky piece, it's not as big a deal.

client/src/pages/RootWorkspacePage/AdminControls/UpdateIsPublicCheckbox.tsx
```diff
@@ -0,0 +1,90 @@
...
+      // it is only here so the user has a moment to see the "Updating"
+      // message if the server responds very quickly
+      setTimeout(() => {
+        this.props.updateWorkspaceIsPublic({
+          variables: {
+            isPublic: !this.props.workspace.isPublic,
+            workspaceId: this.props.workspace.id,
+          }
+        });
+      }, 200);
```
OAGr: Same comment as above

OAGr:
Generally looks good, I encourage you to think a bit more about cleaning up the 200ms part.

client/src/pages/RootWorkspacePage/AdminControls/UpdateIsEligibleCheckbox.tsx
zjmiller: I'll just take this 200ms part out completely. It's only there to ensure that the admin has time to read the word "updating...", but that might not be a big deal.

server/lib/schema/index.ts
zjmiller: I adopted this style just because it had been used throughout the schema. 

I think factoring this out into a separate `requireUser` is a good idea. My preference is to make this a separate refactor PR, since imposing this style uniformly throughout the schema will involve changing many resolvers not related to the admin controls.

client/src/pages/RootWorkspacePage/AdminControls/UpdateIsPublicCheckbox.tsx
zjmiller: I'll take this out too

server/lib/schema/index.ts
OAGr: yep, makes sense

server/lib/schema/index.ts
OAGr: the "separate refactor" argument is a fine one.

PR #215, opened 2018-10-23
Add Linkify plugin

zjmiller:


client/src/lib/slate-plugins/linkifyPlugin.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  change.collapseToEnd();
+}
+
+function LinkifyPlugin() {
+  return {
+    onPaste: (event, change) => {
+      const transfer = getEventTransfer(event);
+      const { type, text } = transfer;
+      if (type !== "text" && type !== "html") {
```
OAGr:
```suggestion
      if (type !== "text" && type !== "html" && !isUrl(text)) {
```

OAGr: The above change would include deleting the bottom segment.

OAGr: I suggest adding some naming to this bit. It's not obvious to me why these conditions work this way.

client/src/lib/slate-plugins/linkifyPlugin.tsx
```diff
@@ -0,0 +1,67 @@
...
+        return;
+      }
+
+      if (!isUrl(text)) {
+        return;
+      }
```
OAGr: I think instead I'd just prefer a boolean for ``isUrl``, and only then do you do the stuff; else you break.

OAGr:
Approved, but you need to fix conflicts.

PR #217, opened 2018-10-23
Staleness

zjmiller:


zjmiller:
Includes the `EpisodeNav` component, which interacts with a workspace's staleness. And includes the the `ContentContainer` component, which just keeps the component structure nice, especially since we want `EpisodeNav` to stretch the width of the screen, but other page content to be contained.

PR #227, opened 2018-10-25
Prettier config

zjmiller:


zjmiller:
Now multiple people will be synced if using Prettier's "format files on save" text editor integration.

Issue #248, opened 2018-11-05
There should be a way to remove pointers with the keyboard


stuhlmueller:
It could be better if the pointer brackets acted more like normal characters, and if deleting a bracket caused the pointer to go away (with the other bracket turned into a plain-text bracket).

The drawback of this approach is that deleting and re-adding a bracket is not a no-op. Because pointers have identity once created, deleting and re-creating will cause the pointer to show up as n/a wherever it is imported.

stuhlmueller:
Also, the "remove pointer" button should be more robust.

stuhlmueller:
One intention behind this change is to make manipulating lists with nested pointers less annoying.

PR #253, opened 2018-11-06
Add pointer removal on Cmd-R

zjmiller:
Make sure your text cursor is in the pointer you'd like to remove. Then press Cmd-R to remove the pointer. (On a PC this might be Ctrl-R? I'll look into this later.)

We can change the keyboard shortcut to whatever we think is best. We can also merge this into the preview branch for experiments.


zjmiller:
Addresses #248

stuhlmueller:
Seems to work! :)

The cursor behavior is a bit funky - when I remove an export, the cursor jumps to the beginning of the block. Is there any way to keep it in the same place?

Cmd-R is already the shortcut for reloading the page, so we can't use that. Not sure what we should use. Maybe ctrl-k (for kill?).

PR #269, opened 2018-11-16
Scheduler changes

zjmiller:


server/test/lib/scheduler/UserSchedule.spec.ts
```diff
@@ -1,86 +1,217 @@
...
+  });
+
+  describe("isActiveInLastWorkspace method", function() {
+    context("with a user that has not been assigned to workspace", function() {
+      it("returns undefined", function() {
+        expect(this.userSchedule.getLastWorkedOnWorkspace()).to.equal(undefined);
+      });
+    });
+
+    context("with a user that has been assigned a workspace and no time has passed", function() {
```
OAGr: I suggest pulling out "with a user that has been assigned a workspace" into a wrapping context (creating an extra level of nesting) to make this more straightforward.

server/test/lib/scheduler/UserSchedule.spec.ts
```diff
@@ -1,86 +1,217 @@
...
+        this.clock.tick(ONE_MINUTE);
+        await this.userSchedule.assignWorkspace(workspaces.get("2-1"));
+        this.clock.tick(ONE_MINUTE);
+        await this.userSchedule.assignWorkspace(workspaces.get("3"));
+        this.clock.tick(ONE_MINUTE);
+        await this.userSchedule.assignWorkspace(workspaces.get("1-1-1"));
+        this.clock.tick(ONE_MINUTE);
+        await this.userSchedule.assignWorkspace(workspaces.get("5-2"));
+        this.clock.tick(ONE_MINUTE);
+        await this.userSchedule.assignWorkspace(workspaces.get("4"));
```
OAGr: this code really could be dryed up a lot. For instance, two of these lines could look like,
```
increment_time(ONE_MINUTE)
await assignWorkspace("4")
```

server/test/lib/scheduler/Schedule.spec.ts
```diff
@@ -100,38 +101,105 @@ describe("Schedule class", function() {
...
+  describe("getTreesWorkedOnLeastRecently", function() {
+    const rootWorkspaces = [
+      workspaces.get("1"),
+      workspaces.get("2"),
+      workspaces.get("3"),
+      workspaces.get("4"),
+      workspaces.get("5"),
+    ];
+
+    it("works in straightforward case", async function() {
```
OAGr: "in a straightforward case" seems like a context.

A more elegant way of doing this would be to have these as contexts, and then have tests for each expect block. ("has the correct length")

server/lib/scheduler/Scheduler.ts
```diff
@@ -20,67 +31,102 @@ class Scheduler {
...
+      if (actionableWorkspaces.length > 0) {
+        return actionableWorkspaces;
+      } else {
+        treesToConsider = _.difference(
+          treesToConsider,
+          [randomlySelectedTree]
+        );
+      }
+    }
+  }
```
OAGr: There must be some better ways of writing the above code, for readability. I'd want something that was more declarative, I found this pretty confusing.

server/lib/scheduler/Scheduler.ts
```diff
@@ -20,67 +31,102 @@ class Scheduler {
...
 
-    let finalWorkspaces = notYetWorkedOn;
+    let eligibleWorkspaces = workspacesWithAtLeastMinBudget;
 
-    // if every workspace in that tree has been worked on
-    // then instead look for workspaces with remaining budgets
-    if (finalWorkspaces.length === 0) {
-      finalWorkspaces = await this.filterByWhetherHasRemainingBudget(eligibleWorkspaces);
-    }
+    const staleWorkspaces = await this.filterByStaleness(workspacesWithAtLeastMinBudget);
```
OAGr: I think there should be some nicer way of doing most of the above, if lines 69-75 only exist for the use of ``staleWorkspaces``. Not sure how to do this with ``await``, but I'd expect something like,
``compose([this.fetchAllWorkspacesInTree, this.filterByWhetherCurrentlyBeingWorkedOn, this.filterByWhetherHasMinBudget, filterByStaleness])`` It really feels like it should be more functional.

server/lib/scheduler/Scheduler.ts
```diff
@@ -20,67 +31,102 @@ class Scheduler {
...
+    return treesWorkedOnLeastRecentlyByUser;
+  }
+
+  private async filterByWhetherCurrentlyBeingWorkedOn(workspaces) {
+    return workspaces.filter(w => !this.schedule.isWorkspaceCurrentlyBeingWorkedOn(w));
+  }
+
+  private filterByWhetherHasMinBudget(workspaces) {
+    return workspaces.filter(w => this.hasMinRemaining(w));
   }
```
OAGr: I like these small functions.

server/lib/scheduler/UserSchedule.ts
```diff
@@ -3,50 +3,84 @@ import { Assignment } from "./Assignment";
...
   }
 
   public getMostRecentAssignment() {
     return this.userSchedule[this.userSchedule.length - 1];
   }
 
+  public getTreesWorkedOnLeastRecentlyByUser(rootWorkspaces) {
+    const treesNotYetWorkedOn = rootWorkspaces.filter(
+      r => this.lastWorkedOnTimestampForTree[r.id] === undefined
+    );
```
OAGr: this feels like it should be a utility method, like,
``public hasNotBeenWorkedOn(workspaceId)``

But maybe in a different file or similar.

server/lib/scheduler/UserSchedule.ts
```diff
@@ -3,50 +3,84 @@ import { Assignment } from "./Assignment";
...
+
+    const lastWorkedOnTimestamps = rootWorkspaces.map(
+      r => this.lastWorkedOnTimestampForTree[r.id]
+    );
+
+    const minTimestamp = Math.min.apply(Math, lastWorkedOnTimestamps);
+
+    const leastRecentlyWorkedOnTrees = rootWorkspaces.filter(
+      r => this.lastWorkedOnTimestampForTree[r.id] === minTimestamp
+    );
```
OAGr: I think I want the above 8 lines to be a new method, like ``leastRecentlyWorkedOnTrees(rootWorskpaces)``

PR #313, opened 2018-12-13
Fixes #47

### LINKED ISSUE #47: Pointer reference open states should be removed from redux after workspaces change
Right now, if you open pointers in one workpace, then click on another workspace with the same blocks as children, the pointers stay open.

coreystaten:
Added a history listener, which closes all pointers during page navigation.

As far as I can tell, it's hard to detect navigation on individual workspace components; hence we just do it globally.  The longer term/future solution is to include the workspace ID as part of the stored pointer state and thread this through to the pointer components, so that expansion can be preserved per workspace.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import * as LogRocket from "logrocket";
-import { BrowserRouter, Route, Redirect, withRouter } from "react-router-dom";
+import { BrowserRouter, Route, Redirect } from "react-router-dom";
 import ApolloClient from "apollo-client";
 import { ApolloProvider } from "react-apollo";
 import { HttpLink } from "apollo-link-http";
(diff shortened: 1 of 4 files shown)


zjmiller:
Hi Corey! A few (minor) comments:

- It might make sense to put the HistoryListenerWithRouter component in a separate file.
- I would probably use `connect` & `mapDispatchToProps` with HistoryListenerWithRouter instead of calling `store.dispatch` directly.
- The following action creator
```
export const closeAllPointerReferences = () => {
  return (dispatch, getState) => {
    dispatch({
      type: CLOSE_ALL_POINTER_REFERENCES
    });
  };
};
```
could just be
```
export const closeAllPointerReferences = () => ({
  type: CLOSE_ALL_POINTER_REFERENCES
});
```
since this isn’t using redux-thunk functionality.
- Using `history.listen` like this means that pointers close when the hash portion of the url changes (I'm pretty sure), and this is something we might not want, though we don’t really use url fragments in the app right now, so this is more a forward-thinking point.

coreystaten:
I've cleaned this up to address some of the issues both of you mentioned:
* It's now in a separate file.
* It uses the standard connect with mapDispatchToProps pattern.
* I found a simpler way to write it without the listen/unlisten pattern, which should be clearer.
* It should now only close pointers on pathname change, not fragment change.

I also renamed it to "ClosePointerListener" since it's likely to only ever do that one job.  I couldn't figure out any more elegant way to do this with react-router; let me know if you have any recommendations.

Issue #307, opened 2018-11-29
Show all kinds of workspace status in tree view


stuhlmueller:
There are a few different kinds of dynamic workspace properties now:

- Is someone actively working on the workspace?
- Is it stale?
- Is it restricted to oracles?

We should show all of these in the tree view.

We should also consider showing budgets in the tree view.

stuhlmueller:
It would be especially nice if workspace status (staleness, restriction to oracles) could also be changed from the tree view.

Mati-Roy:
Is this done? ex.: https://mosaic.ought.org/w/4268/subTree?expanded=true

PR #314, opened 2018-12-14
Adds checkboxes for admin control of workspace status in tree view, #307.

coreystaten:
Added admin checkboxes controlling isStale and isEligibleForOracle flags on workspaces in tree view.  This reuses the checkbox pattern from the admin checkboxes on the root workspace page.  Similar code is now being used in 4 spots, so might want to factor that out into an AdminCheckbox component.


stuhlmueller:
Thanks!

Looking at the tree, there's some redundancy now:

screen shot 2018-12-14 at 7 56 14 am

The indicators and check boxes present the same information. Large trees will be less readable, since the checkboxes take up quite a bit of space.

How difficult would it be to integrate the checkboxes and indicators in a way that lets admins change the status, but also displays the status to non-admins?

coreystaten:
Not difficult, just made a commit to do that.  Current caveat is that "is eligible for oracle" is displaying even when not in oracle mode (as opposed to the "Oracle Only" status display).  We can easily make it not display when not in oracle mode if desired.

coreystaten:
Just noticed that this displays messily when "Was Answered By Oracle" is displayed as well, fixing now.

zjmiller:
Also, as a future PR, I think it would be great to use your approach to refactor the admin controls on the front page: i.e., use `updateQuery` instead of refetching "RootWorkspacesQuery". This will make those checkboxes much more responsive.

Issue #229, opened 2018-10-25
Make homepage look less messy


stuhlmueller:
Factors contributing to messiness:

- Some questions have descriptions (via scratchpad), others don't
- Variable height of questions (depending on scratchpad content, answers)
- Green pointers that stand out

Issue #211, opened 2018-10-21
Speed up front page loading


stuhlmueller:

stuhlmueller:
Especially for admins, but with many trees, it's also slow for normal users.

PR #316, opened 2018-12-14
 Cleans up display of front page.

coreystaten:
Several changes to make the front page display cleaner and more uniform:
* Uses a text display of the main blocks instead of an instance of the slate editor; this allows us to display the blocks (question, answer, and scratchpad/description) on one line each using ellipsis overflow.  It also eliminates noise from green inline pointers.
* Uses flexbox to display editing buttons on the right-hand side above the Tree view button, lowering the vertical display height for admins.

This does have the negative effect of simply showing less info on the front page, since blocks are flattened.  This messes with paragraph structure, pointers (displayed using square brackets), and any other inline elements that might be included in the future.

Currently height of each card is not completely fixed, since the answer display line at the bottom is optional.  This seemed like the best trade-off.

### DIFF
--- a/client/src/lib/slateParser.ts
+++ b/client/src/lib/slateParser.ts
@@ -1,3 +1,4 @@
+import * as _ from "lodash";
 import Plain from "slate-plain-serializer";
 
 export function valueToDatabaseJSON(value: any) {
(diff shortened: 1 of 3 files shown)


coreystaten:
This is work on issue #211

coreystaten:
I had wrongly assumed pointer imports couldn't occur in a root workspace block; this is fixed.  Selectively loading them only when needed might be a good load time improvement as you say -- on the current Mosaic front page, none of the questions I can see make use of imports (unless they're somehow occurring in a root workspace question block?)  If the feature is never used, we could also just disable pointer import in root workspace blocks.

coreystaten:
This also addresses #229

PR #322, opened 2018-12-26
Press enter to autocomplete pointer import

coreystaten:


zjmiller:
I fixed the issue where after pressing enter a new line would briefly appear and disappear. The interaction between "enter" and Slate is kind of confusing. See some discussion here: https://github.com/ianstormtaylor/slate/issues/1345

coreystaten:
Thanks, I figured the behavior was happening inside of Slate, but didn't know that returning false vs undefined from the onKeyDown event would cause the desired behavior by preventing descent down the plugin stack.

Issue #320, opened 2018-12-21
In a subquestion, better visually distinguish between the question and the answer 


zjmiller:

zjmiller:
Adressed in #324

PR #324, opened 2018-12-28
Distinguish question and answer display in episode subquestions using…

coreystaten:
… bullets.  Re: issue #320.


zjmiller:
lgtm

Thanks for abstracting out the BlockBullet and BlockBulletLink component. I agree with could probably dry up the css in that file. But, given that the file is only 30-40 lines of easy-to-follow css, that doesn't seem like a high-priority refactor to me.

PR #328, opened 2019-01-02
Add cypress testing

zjmiller:


zjmiller:
I used async/await whenever I could, but the `auth0-js` API uses callbacks (and doesn't support promises), so in the few files that had to interact directly with this API you'll see slightly more cumbersome syntax.

zjmiller:
After merging these tests, I'll consolidate the auth code that I wrote here with auth code actually in the app. This will involve additional refactoring (e.g., storing all auth-related `localStorage` information within a single data item).

zjmiller:
To actually run these tests yourself, you'll need a copy of `cypress.env.json` which includes auth0 credentials and isn't checked in for that reason. Soon we'll get it set up with Travis CI or CircleCI (#4).

PR #330, opened 2019-01-04
Fixes #257 by refactoring import autocomplete/conversion code.

### LINKED ISSUE #257: When text with imported pointers is pasted, $x should automatically be turned into blue reference

coreystaten:
This adds import conversion on paste, as well as conversion on changing focus away from editor.

Some remaining deficiencies:
* When pasting imports, the cursor ends up after the last import converted instead of at the end of the pasted text.
* Clicking elsewhere in the same editor doesn't convert imports.

### DIFF
--- a/client/src/components/BlockEditor/BlockEditorEditing.tsx
+++ b/client/src/components/BlockEditor/BlockEditorEditing.tsx
@@ -239,15 +239,10 @@ export class BlockEditorEditingPresentational extends React.Component<
       locations = _.filter(locations, loc => !this.isActiveImportPatternLocation(loc));
     }
 
-    let anyConverted = false;
-    locations.forEach(location => {
-      if (this.convertImportPatternLocation(location)) {
-        anyConverted = true;
-      }
-    });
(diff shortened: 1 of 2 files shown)


zjmiller:
I think one refactor that makes sense here is to move as much as possible into a separate file, or separate files.

zjmiller:
One more thing, it would be nice if this worked with submitting new subquestions as well. For example, if a user types "Analyze $1" in the new subquestion form and then clicks Submit. This currently doesn't appear to be working in this PR. I checked this branch out locally and was able to get it to work by wrapping the handler associated with submitting a subquestion in `setTimeout`. Maybe there's a more elegant way to do this.

coreystaten:
Fixed the bug you pointed out, refactored the code into a separate file, and added a global import conversion when the editor component mounts.

Re: submitting a new question, the current won't fix wouldn't work if the newly displayed editors were read only (since those wouldn't use the BlockEditorEditing component) --  a solution for that would be slightly more work.

PR #415, opened 2019-04-01
Auto scroll in tree view

zjmiller:
This PR adds the following feature. When an oracle for a workspace clicks on the "Entire Tree" nav button, the tree that is opened will automatically scroll down to that workspace. Also, the workspace is now highlighted so there's a visual indication it's the active workspace.

### DIFF
--- a/client/src/components/WorkspaceCard/WorkspaceCard.tsx
+++ b/client/src/components/WorkspaceCard/WorkspaceCard.tsx
@@ -52,7 +52,7 @@ const CardBody: any = styled.div`
       "3px solid red"
     )
   }
-  box-shadow: ${(props: any) => props.isActive ? "0 0 10px 5px yellow" : undefined}
+  box-shadow: ${(props: any) => props.isActive ? "0 0 10px 5px yellow" : "none"}
   float: left;
   margin-bottom: 1em;
   width: 42em;
--- a/client/src/components/WorkspaceCard/WorkspaceCard.tsx
+++ b/client/src/components/WorkspaceCard/WorkspaceCard.tsx
@@ -132,10 +132,12 @@ export class WorkspaceCardPresentational extends React.PureComponent<WorkspaceCa
   public componentDidMount() {
     const isThisActiveWorkspace = this.props.activeWorkspaceId === this.props.subtreeQuery.workspace.id;
     if (isThisActiveWorkspace) {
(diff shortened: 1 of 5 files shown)


zjmiller:
There's one aspect of this PR that I hope can be improved upon. Currently, if you scroll to the active workspaces while workspaces above it (above it on the tree view page, not above it in the tree structure) are still loading, then, as these workspaces, load the active workspace moves out of the correct position. 

My approach to this issue is to just add a 3 second delay. After the active workspace part of the tree view loads, we wait 3 seconds for additional workspaces to load, and then scroll to the active workspace.

This won't work when the loading takes longer than 3 seconds, and when it takes less than 3 seconds the user is just waiting for no reason. If we knew when the workspaces "above" the active workspace had finished loading, then we could just wait until this occurred before scrolling the active workspace. But this information is non-trivial to obtain, since all the workspaces fetch information from the GraphQL server independently.

client/src/components/WorkspaceCard/WorkspaceCard.tsx
```diff
@@ -51,6 +52,7 @@ const CardBody: any = styled.div`
       "3px solid red"
     )
   }
+  box-shadow: ${(props: any) => props.isActive && "0 0 10px 5px yellow"}
```
cnorthwood: one of the things I spotted in #416 was that newer TypeScript doesn't like this (as `false` isn't a valid value for `box-shadow`), I think using a ternary is more explicit and will make it happier

```suggestion
  box-shadow: ${(props: any) => props.isActive ? "0 0 10px 5px yellow" : undefined}
```

cnorthwood:
I can see the 3 seconds thing as a reasonable trade-off. In the longer term, I wonder as browsers develop scroll anchoring capabilities it might resolve itself straight forwardly and not need any workarounds? https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring

client/src/components/WorkspaceCard/WorkspaceCard.tsx
zjmiller: Interesting, yeah I noticed those changes in your "Update dependencies" PR.

zjmiller:
Great let's stick with the 3s delay and we can revisit it if need be.

PR #416, opened 2019-04-01
Update dependencies

cnorthwood:


zjmiller:
Made a test experiment with one workspace on the review app. Started participating it and after typing a few characters into the workspace's scratchpad app crashed with following. Revisited the page and there wasn't a crash while I didn't do anything (waited maybe 30 seconds or so), but then after I started typing again got same error.

image

zjmiller:
Got this error (many instances all at once) when I tried to export a pointer (which I had to do in stages because of the previous error limiting how many characters I could get in each time).

image

zjmiller:
Was able to make some pointer using `Cmd-e` (square bracket notation always gave me the above error). After navigating to the tree view, exports weren't numbered and the associated imports were showing up as `NA`, which is basically a sign that the availablePointers prop isn't getting the correct value at the top of the tree (root-level GraphQL query gets these as `connectedPointersOfSubtree` and then passes them down.

cnorthwood:
okay, the first issue only appears to crop up in Chrome, it's fine in Firefox, I can replicate the second in both browsers, will dig in to see what's going on

zjmiller:
I'm glad this is forcing us to do some cleaning up of the codebase.

PR #417, opened 2019-04-01
Use NodeJS Auth0 library in server, rather than browser-based one

cnorthwood:


cnorthwood:
will wait for #416 to go first and then will rebase as the package.json/yarn.lock changes will almost certainly cause a merge conflict

PR #418, opened 2019-04-01
Use scheduler, not query param, for if workspace active

zjmiller:
This uses server-side scheduling data, as opposed to a query param, to determine whether a user should see the "active workspace" UI . This also concerns whether "unlocking" an import is permanent: within the "active workspace" UI it is, but otherwise it isn't. Overall, this makes many things more robust, and will also make it easier to share links without sending users to the "active workspace" version of a workspace.

### DIFF
--- a/client/src/components/BlockEditor/index.tsx
+++ b/client/src/components/BlockEditor/index.tsx
@@ -106,6 +106,7 @@ class BlockEditorPresentational extends React.Component<any, any> {
       isInOracleModeAndIsUserOracle: newProps.oracleModeQuery.oracleMode && newProps.isUserOracle,
       isInOracleMode: newProps.oracleModeQuery.oracleMode,
       isUserOracle: newProps.isUserOracle,
+      isActive: newProps.isActive,
     };
     this.setState({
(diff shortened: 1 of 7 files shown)


stuhlmueller:
1. Is it already obvious to users whether they're in active mode or not?
2. Do we want to keep `isolated` separate from `active`?
3. Do we want to keep experiment ID a query parameter?

zjmiller:
1.  It's not super obvious. There's always a navbar underneath the main header when you're in active mode, and it's not there when you're not in active mode. Also, at this point (with the merging of this PR), you'll only ever be in active mode if you get there via experiment flow.

2. I think we could probably get rid of `isolated`, unless someone can think of a situation where `active` and `isolated` come apart.

3. For the short-term we should probably keep the experiment ID as a query param, because it is what persists across the pages (workspace view page, taking a break page, finding next workspace page) that lets the server know what experiment the user is trying to find a workspace for. 

   There's currently no concept of a user being in an experiment at a time, so we need something like this. It's consistent with the current app that workspaces can be in multiple experiments, that experiments and fallbacks can be related to each other in arbitrarily complex ways, and that users can participate in many different experiments at the same time. 

   We might not need all of this functionality, but we probably want some of it, especially for async experiments. We'll need to make sure whatever replacement we want to implement for the experiment ID query param can preserve what we want.

zjmiller:
I'll merge for now since the suggestions are additional bits of functionality that we can add later.

PR #421, opened 2019-04-02
Malicious oracle rework

zjmiller:
1. Root workspace is assigned to a malicious oracle, and now just involves the creation of the question. Honest oracle gets the root workspace's child.

2. Instructions appear outside of the scratchpad, and change depending on whether it is an oracle's first or second+ time assigned to the workspace.

3. Oracles have two buttons they can use to finish working on a workspace: one ("Next workspace") that takes them to their next assignment without reactivating their parent or submitting the answer, and one ("Done & bubble answer up") that reactivates the parent and submits the answer.

### DIFF
--- a/client/src/pages/EpisodeShowPage/index.tsx
+++ b/client/src/pages/EpisodeShowPage/index.tsx
@@ -7,7 +7,7 @@ import gql from "graphql-tag";
 import styled from "styled-components";
 import { graphql } from "react-apollo";
 import { Helmet } from "react-helmet";
-const ReactMarkdown = require("react-markdown");
+import * as ReactMarkdown from "react-markdown";
 import { compose } from "recompose";
 import { Alert, Row, Col, Button } from "react-bootstrap";
 import { connect } from "react-redux";
--- a/client/src/pages/EpisodeShowPage/index.tsx
+++ b/client/src/pages/EpisodeShowPage/index.tsx
@@ -579,6 +579,7 @@ export class WorkspaceView extends React.Component<any, any> {
                           isActive
                           &&
                           <ResponseFooter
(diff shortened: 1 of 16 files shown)


zjmiller:
As far as refactoring goes, in the GraphQL schema, in the resolver of the `message` field of the `workspace` query, there is a fairly long process to find the experiment associated with the workspace. This actually occurs in the resolver for the `currentlyActiveUser` field as well. I think it makes sense to refactor this. But I'd prefer (i) to do this after this Thursday and (ii) make it a separate refactoring PR.

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -7,8 +7,9 @@ import gql from "graphql-tag";
 import styled from "styled-components";
 import { graphql } from "react-apollo";
 import { Helmet } from "react-helmet";
+const ReactMarkdown = require("react-markdown");
```
cnorthwood: Could this be done in ES6 notation for consistency?
```suggestion
import * as ReactMarkdown from "react-markdown";
```

zjmiller: Interesting. I originally had `import ReactMarkdown from "react-markdown"` and my editor complained the module had no default export. Looking at `node_modules` it looks like the react-markdown package uses Typescript's `export =` syntax which is not the same as `export default` syntax. Typescript documentation (https://www.typescriptlang.org/docs/handbook/modules.html) actually advises using `import ReactMarkdown = require("react-markdown")` in this case, which doesn't even look like JS to me and so I went with the `require` syntax which I knew webpack could handle. Nice to see that `import * as` syntax works though. Seems like the best option!

PR #426, opened 2019-04-04
Show instructions to users from experiment-level instructions

cnorthwood:


zjmiller:
Some initial comments:

1. At the DB/Sequelize level, it might make more sense to introduce an `Instruction` model (i.e., with attributes like `instructionType` and `defaultValue`). Then extending the app with new types of instructions would amount to expanding the `instructionType` enum as opposed to adding an attribute to the `Experiment` model. Also, we could modify the default text associated with an instruction type without having to use a migration. 

   This would also simplify other forms of extension: e.g., we could add a `markupType` enum attribute to the `Instruction` model with potential values like `markdown` and `html` if we wanted to allow different experiments to have different forms of markup. This would be useful if later on we want more expressive power than markdown can provide, but we want to retain backwards compatibility with the older markdown instruction sets.

zjmiller:
2. At the GraphQL level, I think it would be cleaner to subsume all of the instruction type fields under one field of `experiment`, i.e., a field like `instructions`. At the query level, this would turn something like this:

```
experiment(id: $id) {
  id
  name
  metadata
  root_instructions
  honest_oracle_instructions
  malicious_oracle_instructions
  returning_root_instructions
  returning_honest_oracle_instructions
  returning_malicious_oracle_instructions
}
```

into this:

```
experiment(id: $id) {
  id
  name
  metadata
  instructions {
    root_instructions
    honest_oracle_instructions
    malicious_oracle_instructions
    returning_root_instructions
    returning_honest_oracle_instructions
    returning_malicious_oracle_instructions
  }
}
```

cnorthwood:
2 makes sense given 1 - the GraphQL/Sequelize mapping would make it work "magically". 1 hopefully will be a straight forward refactor

zjmiller:
3. Minor pointer: it might make more sense to use camelCase throughout instead of snake_case, as this would be more consistent with the codebase, and more consistent (as far as I can tell) with the prevailing conventions surrounding Sequelize and GraphQL. If you're concerned with keeping your Postgres columns in snake_case, you can do this via the Sequelize field attribute as discussed [here](http://docs.sequelizejs.com/manual/models-definition.html) and [here](https://stackoverflow.com/questions/18858337/sequelize-use-camel-case-in-js-but-underscores-in-table-names).

zjmiller:
The default markdown doesn't seem consistent:

Screen Shot 2019-04-05 at 12 48 17 PM

zjmiller:
Instructions for first-time malicious oracle aren't appearing. `workspace.message` is `null` so something on backend.

zjmiller:
One higher-level comment: Because these commits haven't been merged into anything permanent, I would prefer there to be just one migration, instead of two migrations with the first one essentially canceled out by the second. I think it makes sense to clean things up now before these migrations get merged into `master`, applied to the database, and become indelibly part of the app/codebase. I'd probably feel best if this unnecessary migration complexity were removed from the commit history entirely, then either force pushed here, or we could open a fresh PR.

cnorthwood:
sure - the reason there are 2 migrations was I was trying to avoid rebasing or messing with history too much but can do some squashing to make it make sense

PR #427, opened 2019-04-04
refresh access tokens in the background (fix #244)

### LINKED ISSUE #244: People get logged out much more frequently than they should

cnorthwood:


client/src/auth.tsx
```diff
@@ -99,6 +116,11 @@ export class Auth {
       return false;
     }
 
+    // schedule the tokens for renewal if they've not already done so
+    if (!Auth.renewTokenTimer) {
+      Auth.renewTokenTimer = setTimeout(Auth.renewTokens, 900000); // 15 minutes
```
zjmiller: Is this supposed to be `setInterval` or is it just a one-time refresh?

zjmiller: Also, if it is supposed to an interval, make sure to add a `clearInterval` in `logout`.

cnorthwood: the intention was that a refresh would be scheduled every time an auth check occurs (via `isAuthenticated`), but that's probably confusing and an interval makes more sense, so have updated.

cnorthwood:
https://community.auth0.com/t/using-webauth-checksession-getting-400/15925 suggests that "Refresh Token" needs to be enabled under "Advanced Settings > Grant Types" in Auth0, can you check that @zjmiller?

stuhlmueller:
It's enabled:

Screen Shot 2019-04-15 at 3 41 19 PM

zjmiller:
Issue resolved in different way for now: Auth0 used as sign-up, but then hashed access token serves as a password. So users can stay logged in indefinitely.

Issue #387, opened 2019-03-06
Evaluate alternatives for Logrocket


stuhlmueller:
Issues with Logrocket:
- Can't download videos
- Recordings expire quickly
- Difficult to organize videos

Potential alternatives:
- FullStory
- HotJar

cnorthwood:
Neither HotJar nor FullStory allow downloading of recordings, but FullStory has 30 day retention and Hotjar 1 year retention (although only 300 recordings). As far as I can tell only FullStory gives a way of associating sessions with usernames, so I'd recommend that, if the retention is useful

PR #428, opened 2019-04-04
Replace LogRocket with FullStory

cnorthwood:


cnorthwood:
Fixes #387, however requires a FullStory account and snippet to be set up.

stuhlmueller:
This is the snippet:

```
<script>
window['_fs_debug'] = false;
window['_fs_host'] = 'fullstory.com';
window['_fs_org'] = 'KAQDM';
window['_fs_namespace'] = 'FS';
(function(m,n,e,t,l,o,g,y){
    if (e in m) {if(m.console && m.console.log) { m.console.log('FullStory namespace conflict. Please set window["_fs_namespace"].');} return;}
    g=m[e]=function(a,b,s){g.q?g.q.push([a,b,s]):g._api(a,b,s);};g.q=[];
    o=n.createElement(t);o.async=1;o.src='https://'+_fs_host+'/s/fs.js';
    y=n.getElementsByTagName(t)[0];y.parentNode.insertBefore(o,y);
    g.identify=function(i,v,s){g(l,{uid:i},s);if(v)g(l,v,s)};g.setUserVars=function(v,s){g(l,v,s)};g.event=function(i,v,s){g('event',{n:i,p:v},s)};
    g.shutdown=function(){g("rec",!1)};g.restart=function(){g("rec",!0)};
    g.consent=function(a){g("consent",!arguments.length||a)};
    g.identifyAccount=function(i,v){o='account';v=v||{};v.acctId=i;g(o,v)};
    g.clearUserCookie=function(){};
})(window,document,window['_fs_namespace'],'script','user');
</script>
```

cnorthwood:
This seems to work for me on localhost, I suspect it'll be worth you having a look @stuhlmueller using the review app and see if what you get is worth it. One thing I noted is that my ad blocker was blocking FullStory as a tracking app (probably as expected), but I don't think it did so with LogRocket.

stuhlmueller:
I tried FullStory. Seems like a small improvement over LogRocket. It's unfortunate that recordings still expire very quickly and that you can't download them. You can pay for 3 months retention, but it's $399/month, which seems excessive. Maybe we can reach out to them, tell them that we're a nonprofit and describe what we need for our experiments. [This help page](https://help.fullstory.com/using/startup-plans) makes it seem like that could work out.

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -291,13 +290,9 @@ export class WorkspaceView extends React.Component<any, any> {
...
 
     let readOnlyExportedPointers = [];
-    try {
-      readOnlyExportedPointers = _.flatten(
-        allReadOnlyBlocks.map(b => findPointers(b.value))
-      );
-    } catch (err) {
-      LogRocket.captureException(err);
-    }
+    readOnlyExportedPointers = _.flatten(
```
zjmiller: Does FullStory have a way to log errors? This line has unpredictably thrown errors. We want these errors caught because (1) they don't cause any downstream problems and (2) we don't want the app to crash for users during experiments. We do want to log any error here though so we can figure out what's going on.

cnorthwood: hmm, it looks like you can send strings, so not sure if that's going to be enough: https://help.fullstory.com/develop-js/251886

PR #430, opened 2019-04-05
use ts-node to run dev server directly

cnorthwood:
This gives faster server-reload time in dev and also makes the stack traces refer to the original code, not the transpiled ones in dist.


zjmiller:
This disables the current VS Code debugging setup, which lets you set breakpoints and inspect variables and do other debugging tasks directly in the editor. I'll try to get that working again.

cnorthwood:
Good point. Looks like there's an alternate way to run ts-node including debugging: https://github.com/TypeStrong/ts-node/issues/537

cnorthwood:
okay, this appears to work for me in IntelliJ, so fingers crossed VS Code works as before

zjmiller:
When I edit a `.ts` file nodemon isn't restarting for me. Also, sometimes my breakpoints aren't working: I didn't test this for long, but from what I can tell, the breakpoints work for `.js` files and don't work for `.ts` files.

cnorthwood:
don't think there's a better way of achieving this at present whilst maintaining debug capability and fast reloads, so let's close for now

PR #435, opened 2019-04-08
Answer propagation

zjmiller:


client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -534,6 +534,35 @@ export class WorkspaceView extends React.Component<any, any> {
                   <Row>
                     <Col sm={6}>
                       {
+                        Auth.isAuthenticated()
```
cnorthwood: I wonder if this could be a ternary instead, I think it makes it more obvious that it's an if/else for the message, instead of two &&s with opposite conditions

zjmiller: Yeah that's fair. Let's save it for a general cleanup / breaking up of this component, which hopefully we do soon.

PR #442, opened 2019-04-11
Experiment instructions + server-side upgrade & tidy-up

cnorthwood:


cnorthwood:
The review app seems to work for me. It's a bit of a big PR...

One thing I noticed was that there's a table called `WorkspaceStateCache` that exists (and had foreign keys on Events). As far as I can tell, it's not referenced anywhere else in the code, so I'm wondering if it existed previously but no migration was written to remove it.

zjmiller:
Yeah I'm sure that's right. Let's plan on removing it after we get this big PR merged and working.

PR #444, opened 2019-04-11
cache the root workspace for any particular workspace

cnorthwood:


cnorthwood:
this extracts out the functionality to a common helper and then adds some caching to it - doing this as a draft PR for now so it can be reviewed in isolation to make the review easier, but it builds on top of #442 so can't be merged directly into master.

PR #452, opened 2019-04-12
Cache the root workspace for a particular workspace

cnorthwood:


server/migrations/20190410170255-cache-root-workspace.js
```diff
@@ -0,0 +1,58 @@
...
+      `SELECT "id" FROM "Workspaces"`,
+      {
+        type: Sequelize.QueryTypes.SELECT,
+      }
+    );
+
+    await Promise.all(workspaces.map(async ({ id }) => {
+      const rootId = await findRootWorkspaceId(id);
+      await queryInterface.sequelize.query(
+        `UPDATE "Workspaces" SET "rootWorkspaceId" = :rootId`,
```
zjmiller: This needs a `WHERE` clause. Right now every time it's re-updating every workspace.

cnorthwood: 😬 good catch

PR #454, opened 2019-04-12
 Fixes #257 by refactoring import autocomplete/conversion code. 

### LINKED ISSUE #257: When text with imported pointers is pasted, $x should automatically be turned into blue reference

cnorthwood:


cnorthwood:
this is a rebased version of #330

cnorthwood:
something's not quite right with this refactor. If I add an exported pointer to an editor, and then delete the exported pointer, and type some more text, the pointer sometimes reappears

zjmiller:
There's "auto-convert imports" functionality in #459. Doesn't include convert on copy-paste, but would be much easier to add "convert on paste" by extending #459 as opposed to reworking this PR.

PR #455, opened 2019-04-12
Fixes #70 by refetching connectedPointers on block save.

### LINKED ISSUE #70: Refetch pointer values when navigating to parent or child

cnorthwood:


cnorthwood:
this is #327 rebased on master - I'm not 100% sure I understand the original bug (I can't seem to reproduce it from the description in the ticket), but from doing some testing, nothing seems to be broken

PR #459, opened 2019-04-14
Import conversion on blur, unmount, and subquestion submit

zjmiller:


client/src/components/BlockEditor/BlockEditorEditing.tsx
```diff
@@ -37,6 +38,7 @@ const NextWorkspaceBtn = ({ bsStyle, experimentId, label, navHook }: NextWorkspa
 
 const AUTOSAVE_EVERY_N_SECONDS = 3;
 const DOLLAR_NUMBERS_NOT_NUMBER = /\$[0-9]+[^0-9]/;
+const DOLLAR_NUMBERS = /\$[\d]*/g;
```
cnorthwood: is this regex right? It will match a `$` in isolation without numbers following it. Looking at how it's used, I'm not sure that's the right behaviour - I think reusing the existing `DOLLAR_NUMBERS_NOT_NUMBER` (which is probably confusingly named...) regex might be more correct

client/src/components/BlockEditor/helpers/convertImportsForNodes.ts
```diff
@@ -0,0 +1,74 @@
+import * as uuidv1 from "uuid/v1";
+
+const DOLLAR_NUMBERS = /\$[\d]*/g;
```
cnorthwood: same comment re regex above - I wonder if exporting and reusing the same regex might be better than duplication

client/src/components/BlockEditor/BlockEditorEditing.tsx
zjmiller: Ah great catch. Yeah we definitely want `+`, not `*`. We do, however, want this to match something like `$1` occurring at the end of a text field (i.e., with no character after it), and we want it to be a global search. So I'll change it to `/\$[\d]+/g`.

cnorthwood: okay, makes sense. You could probably drop the `[` and `]` too as you're only matching a single character, I think `/\$\d+/g` would be sufficient

PR #462, opened 2019-04-15
Add Travis CI

cnorthwood:


cnorthwood:
the vast majority of these changes are the result of running prettify, the actual changes of note are in the `package.json`s and the `travis.yml` file.

client/package.json
```diff
@@ -76,8 +78,10 @@
     "@types/recompose": "^0.24.5",
     "cypress": "^3.1.4",
     "immutable": "^3.8.2",
+    "prettier": "^1.16.4",
     "prop-types": "^15.6.1",
     "react-scripts-ts": "^2.14.0",
-    "typescript": "^2.6.2"
+    "typescript": "^2.6.2",
+    "wait-on": "^3.2.0"
```
zjmiller: Is `wait-on` just for cypress? Might make sense to take it out.

cnorthwood: It is. I quite like the idea of having all the infrastructure required to run tests in place, even if the tests aren't actually run, so if we do come back to them we don't need to mess about with the surrounding stuff as much.

zjmiller: Sure, sounds good.

PR #468, opened 2019-04-15
Only enable done button when the response block has content

### LINKED ISSUE #460: Disable "Done" button until there's content in the response field

cnorthwood:
fixes #460


cnorthwood:
this has a downside, as the button state depends on the value of the block as recorded in the store - as the updateBlocks action is throttled, there's a lag between typing/removing content and the button state updating

Issue #461, opened 2019-04-15
Ensure archived workspaces don't hold up scheduling


zjmiller:
Currently, there's a bug in the scheduler where a workspace waits until all of its subquestions have been resolved before it's eligible for rescheduling, even if some of its children are archived.

zjmiller:
Addressed in #469

PR #469, opened 2019-04-15
Parent stale if all subq resolved or archived

zjmiller:
Before, archived subquestions that hadn't been resolved were preventing parents from becoming stale again (where stale basically means eligible for rescheduling).


zjmiller:
Addresses #461

Issue #331, opened 2019-01-07
Store tree expansion status in URL


stuhlmueller:
So that we can link to partially or fully expanded trees.

PR #471, opened 2019-04-22
Store expansion status in URL

mmieszek:
Solution for #331 

A new query string parameter `expand` is introduced which stores a comma-separated list of expanded workspace ids.

When a subtree is collapsed, a corresponding workspace id is removed from the URL.

When there are no expanded nodes, `expand` is removed completely from a query string.

URL is of the form:
`http://localhost:3000/workspaces/2487918d-7ddc-488b-937b-e0b90428e088/subtree?expand=e3cbf7d8-d001-4f7b-8f55-a9827da4f6ed,2487918d-7ddc-488b-937b-e0b90428e088`

**Potential improvements:**
- when a nested subtree is collapsed, remove ids of expanded children from the URL, not only the id of a tree that is being collapsed
- fetch a whole subtree with a single graphql query
- shorten ids in the URL. Storing whole UUIDs is not necessary and it might be enough to store 7-8 characters. The same mechanism employs git to shorten commit hashes. This would make the URL much shorter and easier to share.
- Updated `query-string` dependency to the latest version (`6.4.2`) to have an automated conversion to an array with `arrayFormat = "comma"` option . Right now the query string value is converted into an array manually. Upgrade to TypeScript 3 needed as query-string 6.4.2 uses keyworn `unknown`.

### DIFF
--- a/client/src/components/WorkspaceCard/WorkspaceCard.tsx
+++ b/client/src/components/WorkspaceCard/WorkspaceCard.tsx
@@ -5,11 +5,11 @@ import { Link } from "react-router-dom";
 import { scroller, Element } from "react-scroll";
 import { BlockSection } from "./BlockSection";
 import { ChildrenSection } from "./ChildrenSection";
-import _ = require("lodash");
+import * as _ from "lodash";
+import { parse as parseQueryString, stringify as stringifyQueryString } from "query-string";
+
 import { ChildBudgetBadge } from "../ChildBudgetBadge";
-
 import { AdminCheckboxThatTogglesWorkspaceField } from "../AdminCheckboxThatTogglesWorkspaceField";
-
 import { Auth } from "../../auth";
 
 import {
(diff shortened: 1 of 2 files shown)


zjmiller:
Thanks for submitting this! We're in the process of removing UUIDs from the URLs in favor of shorter identifiers. I think it makes sense to hold off on merging this until we've completed that task, and then recast this PR in those terms.

zjmiller:
It would also be nice to have a system that worked with well the "Expand All" functionality currently associated the `expanded` query param being set to `true`. With this PR, if we "Expand All" and then collapse a few subtrees, the resulting state isn't saved in the URL.

brachbach:
@zjmiller what's the current status of this PR?

zjmiller:
We still need to resolve the issues I mentioned in my comments. Doing this hasn't been a priority, less so now that most non-admins use the compact tree view.

brachbach:
cool, sounds good

PR #481, opened 2019-04-24
Update UI state when logout token expires

cnorthwood:


zjmiller:
We should make sure only someone intending to participate in an experiment sees the "you're logged out" warning. I think the straightforward thing to do here is check whether there's an `experiment` query param, which is currently a surefire way to determine whether a user is in the experiment flow (and there's no near-term plan to refactor this away).

PR #482, opened 2019-04-24
Link root workspace to compact tree by default

cnorthwood:


zjmiller:
The compact tree view only works for more recent experiments. Andreas and I were talking about coming up with a versioning system to help here.

zjmiller:
For now, as a workaround, it might make sense to check the `createdAt` for the root workspace, and then use that to determine whether to link to the old tree view or the new compact tree view.

cnorthwood:
what would the cutoff date be? the introduction of the compact subtree, or does it work on earlier experiments than that?

zjmiller:
OK cutoff date is `2019-04-08 20:52:17.841+00`. But even among trees created on or after this date, we need to know whether they're oracle trees or not. Currently, because we lack a taxonomy for trees and/or experiments, we can only use something like duck typing here. I would probably see if the root workspace is eligible for a malicious oracle. 

So, to sum up, if a tree was created after the cutoff date, and has a root workspace that's eligible for a malicious oracle, then associate it with the compact tree view, otherwise associate with the old tree view.

And we should probably note somewhere that this needs to be cleaned up after a good workspace classification scheme is introduced.

zjmiller:
Buttons get smashed if the window's width is reduced.

image

zjmiller:
Actually, I would just remove the button that navigates to the workspace.

zjmiller:
And probably also remove the "Tree" nav button too, just keeping the text link. We can run this change by @stuhlmueller before merging.

stuhlmueller:
UI looks good.

PR #484, opened 2019-04-25
Allow users to register for notifications when a workspace becomes available

cnorthwood:


cnorthwood:
Once this is ready to merge, someone will need to enable SendGrid on the mosaic-prod application (see https://devcenter.heroku.com/articles/sendgrid#provisioning-the-add-on).

cnorthwood:
At present, I've not been able to test the Slack integration, as I don't think I can create a bot API token for the Ought Slack

PR #493, opened 2019-04-30
Prettier entire project

zjmiller:


zjmiller:
I'm running prettier on everything at once so I can enable "prettier on save" and not have to worry about hard-to-read diffs.

Issue #595, opened 2019-07-12
Locked imports are sometimes expanded in snapshot view


zjmiller:
Though they'll still appear as locked when collapsed.

Issue #596, opened 2019-07-12
Prevent snapshot INITIALIZE from getting overwritten


zjmiller:
This happens in cases where the workspace is loaded more than once while the user is operating within a single assignment.

PR #604, opened 2019-07-15
Workspace history fixes

zjmiller:


zjmiller:
Resolves #595, #596, and #607

PR #623, opened 2019-07-24
Limited assignment api

zjmiller:


zjmiller:
PR no longer needed. Important we keep this branch though because the Heroku app for analytics auto-deploys it.

PR #638, opened 2019-07-30
[WIP] Show link to original document in Experts workspace

### LINKED ISSUE #613: Make a type of pointer that can't be unlocked by judges and is only visible to experts
to use for references to texts

would remove the most obvious way for judges to cheat

### LINKED ISSUE #626: Always give experts easy access to the text
Not an issue for malicious expert since they can just get the link from the full tree view

But honest expert is stuck if the judge neglects to pass the pointer link

(Both in theory, and now that each tree has many honest experts and each participant is honest expert on many trees, also in practice.)

Strategically, there's no time when the judge would ever intentionally not provide the link (since the malicious expert can always get it from the full tree view, so not passing it just asymmetrically disfavors the honest expert)

andrewschreiber:
closes #626 
closes #613 
WIP

### DIFF
--- a/client/src/pages/EpisodeShowPage/index.tsx
+++ b/client/src/pages/EpisodeShowPage/index.tsx
@@ -121,6 +121,11 @@ const WORKSPACE_QUERY = gql`
       }
       rootWorkspace {
         id
+        blocks {
+          id
+          value
+          type
+        }
         tree {
           id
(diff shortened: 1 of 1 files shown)


andrewschreiber:
image
Here's the target UI. Thoughts @brachbach ?

PR #640, opened 2019-07-30
Rename Needs More Work button -> Skip Workspace

### LINKED ISSUE #620: Rename "Needs more work" button to something else
Being confused by the name of the button was one of the top complaints mentioned by experiment participants when I asked them for feedback

The button is currently mostly used:
1. in training experiments in the case where someone doesn't need more practice with a particular role/workspace and wants to leave it for someone else
2. when someone feels too contaminated to complete a workspace
3. when someone thinks that one of the other honest/malicious experts on the tree would do a better job on the workspace than they would

I'm thinking good names could be:
1. "Pass off workspace"
2. "Skip workspace"
3. "Save and pass"

"Save and pass" is probably my favorite, followed by "Skip workspace"

andrewschreiber:
closes #620

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,9 @@ cd /data
 yarn test
 ```
 
-
 ## Development
 
 The code is written in [Typescript](https://www.typescriptlang.org/), but much of it is not correctly annotated. We use [Prettier](https://github.com/prettier/prettier) for code formatting.
-
 
 ## Deployment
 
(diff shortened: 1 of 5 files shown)


andrewschreiber:
Hmm seems like heroku autodeploy on PRs is down.
image

Looks like the last time this happened it was an outage on Heroku's side https://status.heroku.com/incidents/1362.

PR #643, opened 2019-07-31
Don't show root scratchpad if default text

zjmiller:


client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -582,12 +582,23 @@ export class WorkspaceView extends React.Component<any, any> {
...
+      typeof rootWorkspaceScratchpadFirstText === "string" &&
+      rootWorkspaceScratchpadFirstText.trim().toLowerCase() ===
+        "root-level scratchpad";
 
     const doesRootWorkspaceScratchpadContainRelevantContent = !(
       doesRootWorkspaceScratchpadValueContainOneNode &&
-      isRootWorkspaceScratchpadValueFirstTextNodeEmpty
+      (isRootWorkspaceScratchpadFirstTextEmpty ||
+        isRootWorkspaceScratchpadFirstTextDefaultValue)
     );
```
andrewschreiber: Although nice and compact, this logic is hard to follow. Maybe something like this instead:
```javascript
doesRootWorkspaceScratchpadContainRelevantContent = !isRootWorkspaceScratchpadFirstTextEmpty && 
                                                    !isRootWorkspaceScratchpadFirstTextDefaultValue && 
                                                    !doesRootWorkspaceScratchpadValueContainOneNode
```

Could you also add a comment to explain the cases where `doesRootWorkspaceScratchpadValueContainOneNode` would be true? Without using Slate extensively, this is unclear.

PR #648, opened 2019-08-01
Correct stale behavior on API-based experiment creation

### LINKED ISSUE #594: API bug: when creating trees without top-level answers provided, top-level honest workspace not stale
Of course, the next workspace that needs work should always be stale. 

Worth testing this for all possible sets of info provided via API -- just top-level Q, also honest top-level answer, also malicious top-level answer

andrewschreiber:
closes #594

Adds await calls, even though it will slightly slow down the script, to guarantee consistent behavior.

### DIFF
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -2069,7 +2069,6 @@ const schema = new GraphQLSchema({
                   },
                 );
 
-                // Make sure appropriate workspaces are stale based on answers provided
                 // Honest child workspace is always created, so top-level question is never stale.
                 await workspace.update({
                   isStale: false,
@@ -2165,7 +2164,6 @@ const schema = new GraphQLSchema({
                       isPublic: false,
                       shouldOverrideToNormalUser: false,
                     });
-
                   }
                 }
 
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -2071,8 +2071,10 @@ const schema = new GraphQLSchema({
                 );
 
                 // Make sure appropriate workspaces are stale based on answers provided
-
-                /* await */ workspace.update({ isStale: false });
+                // Honest child workspace is always created, so top-level question is never stale.
+                await workspace.update({
+                  isStale: false,
+                });
 
                 const tree = await Tree.create({
                   rootWorkspaceId: workspace.id,
@@ -2100,7 +2102,9 @@ const schema = new GraphQLSchema({
                   shouldOverrideToNormalUser: false,
                 });
 
-                /* await */ honestChild.update({ isStale: false });
+                await honestChild.update({
+                  isStale: honestAnswer ? false : true,
+                });
 
                 if (honestAnswer) {
                   const blocks = (await honestChild.$get("blocks")) as Block[];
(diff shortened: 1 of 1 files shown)


server/lib/schema/index.ts
```diff
@@ -2158,6 +2165,10 @@ const schema = new GraphQLSchema({
                       isPublic: false,
                       shouldOverrideToNormalUser: false,
                     });
+
+                    await judgeChild.update({
+                      isStale: maliciousAnswer && honestAnswer ? true : false,
```
zjmiller: If a judge workspace is created using this GraphQL mutation, it will always be the leaf workspace, and so should always be stale. Another way to get a feel for what is going on is notice that this line is within the scope of two conditionals: the outer conditional guaranteeing `honestAnswer` has been provided and the inner one guaranteeing `maliciousAnswer` has been provided, so `maliciousAnswer && honestAnswer` will always be true.

andrewschreiber: Good catch.

It seems Mosaic automatically does workspace creation / staleness adjustment outside this function. I'm trying to figure out where that happens, as the above logic doesn't result in the behavior that I'd expect.

Also planning to set up a CSV file that is structured to test all the cases we enumerated.

zjmiller: Mosaic, when creating a workspace, always sets `isStale` to true. It's true though that there is a lot of logic, spread throughout the app, that can then set `isStale` to false (indicating that the workspace shouldn't be scheduled again unless something significant happens, like all of its subquestions have answers) or set it to true again (indicating the workspace should be scheduled again, all else equal). 

None of this complexity is that relevant here though. This API is used to do the first few workspaces of a tree ahead of time, with some flexibility as to exactly how many to do. We need to make sure that the leaf workspace, whichever it is, is eligible to get scheduled to actual experiment participants (i.e., is marked as stale) and the non-leaf workspaces aren't eligible (since they've already been handled by the user of the API).

zjmiller: But yeah it is important to know that the default value for `isStale` is true. You can see this in the workspace model.

So for this PR we just need to set `isStale` as false for any non-leaf workspace.

andrewschreiber:
On staleness w.r.t https://github.com/oughtinc/mosaic/issues/594#issuecomment-517810059

Top-level Q only ✅
-> Workspace 50 should be stale
image

Top-level Q + honest answer✅
image

Top-level Q + honest answer + malicious answer ✅
image

server/lib/schema/index.ts
```diff
@@ -297,7 +297,6 @@ export const workspaceType = makeObjectType(
 
 import { UserActivityType } from "./UserActivity";
 import { WorkspaceActivityType } from "./WorkspaceActivity";
-import { generateMaliciousQuestionValue } from "../models/helpers/generateMaliciousQuestionValue";
```
zjmiller: Nice catch. Kinda surprised Travis CI didn't notice this? If you wanted to take a look at the Travis CI setup, or recommend any other kind of tooling like some sort of automated linting, I think that would be valuable.

andrewschreiber: Hmm haven't looked into automated linting outside of my IDE. Could be interesting to take a look next week, time permitting

zjmiller: Cool, no rush. Something to keep in mind when we're thinking about next task to add to queue though.

PR #650, opened 2019-08-03
Most interesting branch (w/o restarts)

zjmiller:
So far have implemented:

- Trees now by default come with a `isMIBWithoutRestarts` field set to false
- Any tree with this field set to false (or null as will be the case for older trees) will be the same as before
- Any tree with this field set to true will exhibit MIB w/o restarts behavior
  - If a malicious oracle chooses to challenge, it will create a workspace that starts off as honest
  - An honest expert assigned to such a workspace will be given a choice: (1) Concede to malicious or (2) Play out this subtree
  - If the honest expert picks (1), the malicious answer will be passed to the upstream workspace
  - If the honest expert picks (2), the workspace transforms into a normal judge workspace
  - If a judge selects either A1 or A2 as their answer, the upstream judge is not reactivated for scheduling
  - A judge can only ask one subquestion per assignment (after asking one question redirects to "Break Between Assignments" page)
  - The first malicious expert workspace after the root isn't allowed to decline to challenge
  - The first honest expert deciding whether to concede to malicious or play out the subtree isn't allowed to concede to malicious
- Compact tree view should be compatible with changes
- Experiment API can be used to create tree with `isMIBWithoutRestarts` set to true
- Admin UI can toggle `isMIBWithoutRestarts`


server/lib/schema/helpers/extractMaliciousAnswerValueFromMaliciousQuestionAndJudgeQuestion.ts
```diff
@@ -0,0 +1,49 @@
+const uuidv4 = require("uuid/v4");
+
+export function extractMaliciousAnswerValueFromMaliciousQuestionAndJudgeQuestion(
+  maliciousQuestionValue,
+  judgeQuestionValue,
+) {
+  const honestPointerId = maliciousQuestionValue[0].nodes[3].data.pointerId;
+  const firstPointerId = judgeQuestionValue[0].nodes[3].data.pointerId;
+  const secondPointerId = judgeQuestionValue[0].nodes[5].data.pointerId;
```
andrewschreiber: Feeling a little concerned about these index references (`nodes[5]`). 

How confident are you that these indexes will remain fixed? Maybe add some checks here.

Also would appreciate a comment in the code documenting why you are accessing these particular indexes

server/lib/schema/index.ts
```diff
@@ -1291,37 +1434,46 @@ const schema = new GraphQLSchema({
...
+                            );
                           }
-                          if (allResolvedOrArchived) {
-                            await greatGrandparentWorkspace.update({
-                              isStale: true,
-                              isNotStaleRelativeToUser: [],
-                            });
+                          const isNotRoot = greatGrandparentWorkspace.parentId;
+                          if (isNotRoot) {
+                            const children = (await greatGrandparentWorkspace.$get(
```
andrewschreiber: I'm confused by the naming here as it sounds like a Bool: `isNotRoot`.

client/src/components/NewBlockForm.tsx
```diff
@@ -87,6 +100,9 @@ export class NewBlockFormPresentational extends React.Component<any, any> {
             : 1,
         }}
       >
+        {this.state.shouldRedirectToBreakPage && (
+          <Redirect to={`/break?e=${this.props.experimentId}`} />
+        )}
```
andrewschreiber: Haven't looked into the object lifecycle here, but noticing that `shouldRedirectToBreakPage` isn't set back to `false`. Is that okay?

client/src/pages/CompactTreeView/MaliciousAnswer.tsx
```diff
@@ -21,48 +21,54 @@ export class MaliciousAnswer extends React.PureComponent<any, any> {
...
+            </CompactTreeRowContent>
+          ) : (
+            <span style={{ color: "#999" }}>Waiting for response</span>
+          )}
+        </CompactTreeRow>
+        {didHonestDecideToConcede && (
+          <div style={{ color: "#777" }}>Honest conceded</div>
         )}
-      </CompactTreeRow>
+      </React.Fragment>
```
andrewschreiber: Noticing you're using `React.Fragment` here instead of `div`. Does your motivation match https://reactjs.org/docs/fragments.html?

client/src/pages/CompactTreeView/V2CompactTreeGroupContainer.tsx
```diff
@@ -107,15 +107,28 @@ export class CompactTreeGroupContainerBase extends React.PureComponent<
...
           isSamePointerInHonestAnswerDraftAndMaliciousQuestion);
 
       const didMaliciousWin =
-        !didHonestWin && // this is here b/c in some older trees the follow two conjuncts are true in cases where the malicious oracle declines
-        isHonestOracleCurrentlyResolved &&
-        !isSamePointerInHonestAnswerDraftAndMaliciousQuestion;
+        (!didHonestWin && // this is here b/c in some older trees the follow two conjuncts are true in cases where the malicious oracle declines
+          isHonestOracleCurrentlyResolved &&
+          !isSamePointerInHonestAnswerDraftAndMaliciousQuestion) ||
+        didHonestDecideToConcede;
```
andrewschreiber: The win logic is complex, but it seems fair as the underlying system is complex. I'd like to detect a situation where both `didHonestWin` and `didMaliciousWin` are `true`. Could we add a check here?

client/src/pages/EpisodeShowPage/EpisodeNav.tsx
```diff
@@ -131,7 +131,7 @@ class EpisodeNavPresentational extends React.Component<EpisodeNavProps, any> {
               <TakeBreakBtn
                 bsStyle="primary"
                 experimentId={experimentId}
-                label="Skip Workspace"
+                label="Skip workspace"
```
andrewschreiber: Better UI one character at a time :)

client/src/pages/EpisodeShowPage/WorkspaceRelations.ts
```diff
@@ -34,19 +32,12 @@ const WORKSPACE = "WORKSPACE";
...
-
 const RelationTypeAttributes = [
   {
     name: WorkspaceRelationTypes.WorkspaceQuestion,
     source: WORKSPACE,
     blockType: QUESTION,
-    permission:
-      Auth.isAdmin() && !isUserInExperiment
-        ? Permissions.Editable
-        : Permissions.ReadOnly,
```
andrewschreiber: Looks like this removes the ability of admins to edit the Question. Was this your intention?

server/lib/models/workspace.ts
```diff
@@ -494,27 +499,52 @@ export default class Workspace extends Model<Workspace> {
...
+      return true;
+    }
+
+    // If parent is non-expert, then honest
+    if (!isParentMaliciousOracleWorkspace && !isParentHonestOracleWorkspace) {
+      return true;
+    }
+
+    // Else not honest
+    return false;
```
andrewschreiber: @brachbach Does this logic match up with your expectations for honest children?

server/lib/models/workspace.ts
```diff
@@ -557,22 +587,42 @@ export default class Workspace extends Model<Workspace> {
     isEligibleForMaliciousOracle,
     shouldOverrideToNormalUser,
   }) {
+    let isAwaitingHonestExpertDecision;
```
andrewschreiber: If `shouldOverrideToNormalUser` is false and `isMIBWithoutRestarts` is true, `isAwaitingHonestExpertDecision` is not assigned a value before being passed to `Workspace.create`. Is there a default value you'd rather instead or a different logic tree?

server/lib/schema/helpers/extractMaliciousAnswerValueFromMaliciousQuestionAndJudgeQuestion.ts
```diff
@@ -0,0 +1,49 @@
...
+      type: "line",
+      isVoid: false,
+      data: {},
+      nodes: [
+        {
+          object: "text",
+          leaves: [
+            {
+              object: "leaf",
+              text: "A ",
```
andrewschreiber: Is this `A` for answer?

client/src/pages/EpisodeShowPage/HonestExpertDecisionBtns.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {!isGreatGrandparentRootWorkspace && (
+          <TakeBreakBtn
+            bsStyle="primary"
+            experimentId={experimentId}
+            label="Concede to malicious expert"
+            navHook={() => {
+              snapshot("CONCEDE_TO_MALICIOUS");
+              if (concedeToMalicious) {
+                markAsNotStale();
+                concedeToMalicious();
```
andrewschreiber: I think `concedeToMalicious` expects a workspace `id` passed in as an argument. Is this happening implicitly?

server/lib/schema/index.ts
```diff
@@ -1159,6 +1180,124 @@ const schema = new GraphQLSchema({
...
+            if (grandParentWorkspace === null) {
+              return false;
+            }
+
+            const judgeBlocks = (await workspace.$get("blocks")) as Block[];
+            const judgeQuestion = judgeBlocks.find(b => b.type === "QUESTION");
+
+            const maliciousBlocks = (await parentWorkspace.$get(
+              "blocks",
+            )) as Block[];
```
andrewschreiber: Thanks for adding types here and elsewhere in this PR! :D

server/lib/schema/index.ts
```diff
@@ -1159,6 +1180,124 @@ const schema = new GraphQLSchema({
...
+              );
+              if (upstreamJudge === null) {
+                return false;
+              }
+              const children = (await upstreamJudge.$get(
+                "childWorkspaces",
+              )) as Workspace[];
+              let allResolvedOrArchived = true;
+              for (const child of children) {
+                if (!(child.isCurrentlyResolved || child.isArchived)) {
```
andrewschreiber: Would prefer `if (!child.isCurrentlyResolved  || !child.isArchived)` to reduce mental gymnastics (and save 1 character ;) )

zjmiller:
Thanks Andrew! As far as testing goes, anything listed in the very first comment of this PR would be good to check out.

server/lib/schema/helpers/extractMaliciousAnswerValueFromMaliciousQuestionAndJudgeQuestion.ts
zjmiller: The malicious and judge question values are both constructed programmatically, so these index references will be OK as long as that question value generation code stays the same. 

I agree it's a form of coupling that doesn't feel great. I'll add some checks and leave a comment.

server/lib/schema/index.ts
zjmiller: The thought is that if `parentId` is not undefined, then `isNotRoot` will be truthy. I think I'll change it to 

`const isRoot = _.isNil(greatGrandparentWorkspace.parentId);` 

and 

`if (!isRoot)`.

client/src/components/NewBlockForm.tsx
zjmiller: Yeah this should be fine. There are examples set up in exactly this way, and I assume `<Redirect />` unmounts the original component, resetting the state. It's unclear how you could even set it back to false.

client/src/pages/CompactTreeView/MaliciousAnswer.tsx
zjmiller: Yeah exactly. It's not a big deal here, but in general I try to avoid using a `<div>` solely as a workaround for the fact React won't let you just use an unwrapped sequence of components..

client/src/pages/CompactTreeView/V2CompactTreeGroupContainer.tsx
zjmiller: Sure

client/src/pages/EpisodeShowPage/WorkspaceRelations.ts
zjmiller: Yeah. Sometimes admins participate in experiments with their admin accounts (maybe we should discourage this, but it's a separate point). And there's now an edge case situation (when an honest expert admin is viewing the "original judge workspace" when deciding whether to concede to malicious or play out subtree) where we want admins in this situation to see the workspace question in the same way a non-admin would.

server/lib/models/workspace.ts
zjmiller: I think `isAwaitingHonestExpertDecision` always get a value here. 
- If `shouldOverrideToNormalUser` is true, then it gets set to false on line 595
- If `shouldOverrideToNormalUser` is false, then
  - If `isMIBWithoutRestarts` is true, then it gets set to a dynamically calculated value on line 617.
  - If `isMIBWithoutRestarts` is false, then it gets set to false on line 624.

Can you double check this?

server/lib/schema/helpers/extractMaliciousAnswerValueFromMaliciousQuestionAndJudgeQuestion.ts
zjmiller: Yeah exactly.

server/lib/schema/index.ts
zjmiller: Sure. If you distribute `!` over disjunction though it turns into conjunction, so will be:

`if (!child.isCurrentlyResolved && !child.isArchived)`

client/src/pages/EpisodeShowPage/HonestExpertDecisionBtns.tsx
zjmiller: This is the concedeToMalicious prop in the parent component: 
```js
concedeToMalicious={() =>
  this.props.concedeToMalicious({
    variables: { id: workspace.id },
  })
}
```

client/src/pages/EpisodeShowPage/WorkspaceRelations.ts
brachbach: @zjmiller FWIW I think it'd be fine to tell admins not to participate in experiments with their admin accounts

I think it'd be helpful if you wrote up a short guide for people who have an admin account to:
- participate in an experiment
- make changes to the tree via the UI
- see workspaces just like the participant in the workspace would have seen it (to the extent that's possible)

and it'd be fine if for some of these the best thing is to use a non-admin account

server/lib/models/workspace.ts
brachbach: Yep, though I'll admit I hadn't thought it through before and I now see that MIB will make the subtree view even weirder :p

server/lib/models/workspace.ts
andrewschreiber: Looks good, I missed the `isMIBWithoutRestarts` is true case. Thanks for the line numbers, made it easy to re-check!

server/lib/schema/index.ts
andrewschreiber: Good catch. Noticed this is unchanged thus far, but I marked the PR as LGTM because I figure this is a quick change that we don't need to go back-and-forth on.

andrewschreiber: Never mind about unchanged, it turns out that going through `View changes` shows the outdated version of the code.
image

PR #669, opened 2019-08-16
Add Heap

stuhlmueller:
So that we can more easily track what's happening without writing custom analytics code.


stuhlmueller:
If you want to make this PR better, add [`heap.identify('[username goes here]');`](https://docs.heap.io/docs/using-identify#section-when-and-where-to-call-identify-and-adduserproperties) in the right place?

stuhlmueller:
Users are not yet identified on heap, not sure why.

PR #673, opened 2019-08-19
Mark honest/malicious pointers during honest decision

zjmiller:


zjmiller:
image

zjmiller:
image

zjmiller:
image

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -282,7 +317,11 @@ class PointerImportNodePresentational extends React.Component<any, any> {
         transition: "background-color 0.4s",
         whiteSpace: "nowrap",
         ":hover": {
-          backgroundColor: isLazyPointer
+          backgroundColor: isKnownAsMaliciousAnswerCandidate
+            ? "red"
+            : isKnownAsHonestAnswerCandidate
+            ? "green"
+            : isLazyPointer
             ? "red"
```
andrewschreiber: Noticing the same `red` is being used for `isKnownAsMaliciousAnswerCandidate` and `isLazyPointer` here (and above). I think its good practice to use different colors for each state if possible.

@brachbach do you have any color preferences here?

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -213,8 +231,18 @@ class PointerImportNodePresentational extends React.Component<any, any> {
...
+      isAwaitingHonestExpertDecision,
     } = this.props;
 
+    const exportPointerId = this.props.nodeAsJson.data.pointerId;
+    const isKnownAsHonestAnswerCandidate =
+      isAwaitingHonestExpertDecision &&
+      idOfHonestAnswerCandidate === exportPointerId;
+    const isKnownAsMaliciousAnswerCandidate =
+      isAwaitingHonestExpertDecision &&
+      idOfMaliciousAnswerCandidate === exportPointerId;
```
andrewschreiber: nitpick: swap the order of 240 & 243, so it matches the logic chain order below.

client/src/lib/slate-pointers/PointerImportNode.tsx
```diff
@@ -258,7 +285,11 @@ class PointerImportNodePresentational extends React.Component<any, any> {
     const styles = StyleSheet.create({
       OuterPointerImportStyle: {
         ":before": {
-          backgroundColor: isLazyPointer
+          backgroundColor: isKnownAsMaliciousAnswerCandidate
+            ? "red"
+            : isKnownAsHonestAnswerCandidate
+            ? "green"
+            : isLazyPointer
```
andrewschreiber: I see use of both `isLazyPointer` and `isLazy` for this same logic. Do you think we could use the same name or is there a good reason for them to be named differently?

server/lib/models/workspace.ts
```diff
@@ -189,6 +189,95 @@ export default class Workspace extends Model<Workspace> {
...
+
+      const idOfA1AnswerCandidate = _.get(
+        judgeQuestionBlock,
+        "value[0].nodes[3].data.pointerId",
+      );
+
+      const idOfA2AnswerCandidate = _.get(
+        judgeQuestionBlock,
+        "value[0].nodes[5].data.pointerId",
+      );
```
andrewschreiber: Need some checks for these. Your approach in the main PR works great

server/lib/models/workspace.ts
```diff
@@ -189,6 +189,95 @@ export default class Workspace extends Model<Workspace> {
...
+      if (isExpertWorkspace) {
+        return;
+      }
+
+      // The pointer containing the malicious answer candidate
+      // first appears in the judge question. But to distinguish
+      // it from the honest answer candidate, we need to know which pointer
+      // contains the honest answer candidate, and this first appears in the
+      // malicious question.
+      const maliciousWorkspace = await Workspace.findByPk(workspace.parentId);
```
andrewschreiber: Great comment, thanks!

client/src/lib/slate-pointers/PointerImportNode.tsx
zjmiller: We don't really use lazy pointers at the moment, and the context would presumably disambiguate for the user. But curious what @brachbach thinks.

PR #677, opened 2019-08-21
Add sketch of assignment scoring

brachbach:
DO NOT MERGE -- is just to have a place to put the assignment scoring code forhttps://docs.google.com/document/d/1oj19c1dJcz4Xeu5CdYExpCEzINy86gDtbsx8oPOo3uA/edit#

### DIFF
--- a/assignmentScoring/assignmentScoring.ts
+++ b/assignmentScoring/assignmentScoring.ts
@@ -1,0 +1,97 @@
+enum Role {
+    Judge = 0,
+    Honest,
+    Malicious
+};
+
+type workspace = {
+    role: Role;
+};
+
+type participant = {};
+
+type tree = {
+    // priority is unique and highest priority is 1 -- so there's a tree with priority 1, then with priority 2, etc.
+    priority: number;
+};
+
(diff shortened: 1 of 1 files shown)


assignmentScoring/assignmentScoring.ts
```diff
@@ -0,0 +1,97 @@
+enum Role {
```
brachbach: FYI I'm still planning to improve this code as part of thinking through assignment prioritization, sharing it now so that Zak can give a rough estimate of dev hours to implement something like this.

I do think it's far enough along that feedback about the proposed scoring rules would be helpful

I'd say don't worry too much about the code quality, we're not planning to actually use this code :)

assignmentScoring/assignmentScoring.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+type participant = {};
+
+type tree = {
+    // priority is unique and highest priority is 1 -- so there's a tree with priority 1, then with priority 2, etc.
+    priority: number;
+};
+
+const mockGetTreeThatContainsWorkspace = (workspace: workspace, priorityForMock: number): tree => {
```
zjmiller: What's the `priorityForMock` parameter doing here?

PR #679, opened 2019-08-23
Remove find suboptimal workspace button

### LINKED ISSUE #667: Remove "Find Suboptimal Workspace" button for now
It currently doesn't do anything:

```

Ben Rachbach  17 minutes ago
@Zachary Miller is it true that the “Find Suboptimal Workspace” button currently doesn’t do anything, because in our current FE setup there are no workspaces that’d ever fall into the the “suboptimal” category?

Zachary Miller  14 minutes ago
basically yeah, the only case where it matters is you could be assigned to both a parent judge workspace and its child judge-to-judge workspace

```

And is rather confusing

LMK if this would take more than an hour of dev time, I assumed not since it's just removing something

andrewschreiber:
closes #667


andrewschreiber:
Reduced box height from 200px to 100px.

PR #680, opened 2019-08-23
Confirm that a newly asked subquestion is in Redux before snapshot

zjmiller:


client/src/components/NewBlockForm.tsx
```diff
@@ -65,8 +65,11 @@ export class NewBlockFormPresentational extends React.Component<any, any> {
     };
   }
 
-  public componentWillReceiveProps() {
-    if (this.state.pending) {
+  public componentWillReceiveProps(newProps) {
+    if (
+      this.state.pending &&
+      newProps.numBlocksInRedux > this.props.numBlocksInRedux // This checks whether the newly asked sub-question has made it into Redux
```
andrewschreiber: This seems fine as long as Redux will trigger a new props refresh, which it should.

Do you think the `setState` on 81 should be moved out of this block?

client/src/components/NewBlockForm.tsx
```diff
@@ -388,11 +391,15 @@ export class NewBlockFormPresentational extends React.Component<any, any> {
...
+const mapStateToProps = state => ({
+  numBlocksInRedux: state.blocks.blocks.length,
+});
+
 const mapDispatchToProps = dispatch => ({
   resetBlock: ({ id }) => dispatch(resetBlock({ id })),
 });
 
 export const NewBlockForm = connect(
-  undefined,
```
andrewschreiber: Curious how we think about `undefined` versus `null`. Notice we use `null` ~284 times and `undefined` ~33 times.

zjmiller: Yeah good question... Different people having different preferences. Might be something to keep an eye on in future PRs. Let me know if you have a preference for consistency.

andrewschreiber: Looked into this a bit. I think we should use `null`.
image
It seems in idiomatic javascript, when assigning an empty property value, null is preferred over undefined.
https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

andrewschreiber:
Do you want a re-review here?

PR #701, opened 2019-08-30
Update Existing Trees (for API)

### LINKED ISSUE #572: Issues related to API (post- release of API v3)
- set whether judge-to-judge questions are allowed via the API (or default to not allowing them)
- change expert assignments on existing trees via the API
- remove trees from an experiment via the API
- set experiment as active when creating it via the API (or experiments default to being active when created)
- set experiment metadata via the API
- allow pointers to be passed down from one workspace created via the API to another (not super important, probably not worth the dev effort)
- return the IDs of the created trees (not super important)

andrewschreiber:
closes #572

see https://github.com/oughtinc/mosaic-api-scripts/pull/16

### DIFF
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -2204,13 +2204,10 @@ const schema = new GraphQLSchema({
               await transaction.commit();
               return true;
             } catch (err) {
-              if (err) {
-                console.error("updateTreeExperts error", err);
-                await transaction.rollback();
-                return false;
-              }
+              console.error("updateTreeExperts error", err);
+              await transaction.rollback();
+              return false;
             }
-            return false;
           },
         ),
       },
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -43,7 +43,7 @@ import NotificationRequest from "../models/notificationRequest";
 import { generateJudgeQuestionValue } from "../models/helpers/generateJudgeQuestionValue";
 import { generateMaliciousQuestionValue } from "../models/helpers/generateMaliciousQuestionValue";
 
-import * as sequelize from "../models/index";
+import * as models from "../models/index";
 
 const generateReferences = references => {
   const all = {};
@@ -402,7 +402,7 @@ const TreeReplaceExpertInput = new GraphQLInputObjectType({
   name: "treeReplaceExpertInput",
   fields: {
     treeId: { type: GraphQLString },
-    replaceOracles: { type: GraphQLBoolean },
+    shouldReplaceOracles: { type: GraphQLBoolean },
     emailsOfHonestOracles: { type: GraphQLList(GraphQLString) },
     emailsOfMaliciousOracles: { type: GraphQLList(GraphQLString) },
   },
@@ -2135,13 +2135,13 @@ const schema = new GraphQLSchema({
           async (_, { bulkTreeInputs }, context) => {
             let transaction;
             try {
-              transaction = await sequelize.default.transaction();
+              transaction = await models.default.transaction();
               for (const treeInput of bulkTreeInputs) {
                 const {
                   treeId,
                   emailsOfHonestOracles,
                   emailsOfMaliciousOracles,
-                  replaceOracles,
+                  shouldReplaceOracles,
                 } = treeInput;
 
                 const tree: Tree = await Tree.findByPk(treeId);
@@ -2149,7 +2149,7 @@ const schema = new GraphQLSchema({
                   throw new Error(`No tree found for TreeId ${treeId}`);
                 }
 
-                if (replaceOracles) {
+                if (shouldReplaceOracles) {
                   await UserTreeOracleRelation.destroy({
                     where: {
                       TreeId: tree.id,
@@ -2210,9 +2210,8 @@ const schema = new GraphQLSchema({
               await transaction.commit();
               return true;
             } catch (err) {
-              console.error("updateTreeExperts error", err);
               await transaction.rollback();
-              return false;
+              throw err;
             }
           },
         ),
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -2098,29 +2098,55 @@ const schema = new GraphQLSchema({
         type: GraphQLBoolean,
         args: {
           treeId: { type: GraphQLString },
-          honest1Email: { type: GraphQLString },
-          honest2Email: { type: GraphQLString },
-          maliciousEmail: { type: GraphQLString },
+          honestEmails: { type: GraphQLList(GraphQLString) },
+          maliciousEmails: { type: GraphQLList(GraphQLString) },
         },
         resolve: requireAdmin(
           "You must be logged in as an admin to update experiment expert assignments",
-          async (
-            _,
-            { treeId, honest1Email, honest2Email, maliciousEmail },
-            context,
-          ) => {
+          async (_, { treeId, honestEmails, maliciousEmails }, context) => {
             const tree: Tree = await Tree.findByPk(treeId);
             if (tree === null) {
               return false;
             }
-            const honest1User = await User.userOrNullForEmail(honest1Email);
-            const honest2User = await User.userOrNullForEmail(honest2Email);
-            const maliciousUser = await User.userOrNullForEmail(maliciousEmail);
-            oracleRelations = [];
 
-            await tree.update({
-              oracleRelations,
+            await UserTreeOracleRelation.destroy({
+              where: {
+                TreeId: tree.id,
+              },
             });
+
+            for (const emailOfHonestOracle of honestEmails) {
+              const user = await User.findOne({
+                where: {
+                  email: emailOfHonestOracle,
+                },
+              });
+
+              await tree.$add("oracle", user);
+            }
+
+            for (const emailOfMaliciousOracle of maliciousEmails) {
+              const user = await User.findOne({
+                where: {
+                  email: emailOfMaliciousOracle,
+                },
+              });
+
+              await tree.$add("oracle", user);
+
+              const oracleRelation = await UserTreeOracleRelation.findOne({
+                where: {
+                  UserId: user.id,
+                  TreeId: tree.id,
+                },
+              });
+
+              if (oracleRelation === null) {
+                return false;
+              }
+
+              await oracleRelation.update({ isMalicious: true });
+            }
 
             return true;
           },
--- a/server/lib/schema/index.ts
+++ b/server/lib/schema/index.ts
@@ -2094,27 +2094,38 @@ const schema = new GraphQLSchema({
           },
         ),
       },
-      updateTreeExpertAssignments: { 
+      updateTreeExpertAssignments: {
         type: GraphQLBoolean,
-        args: { // TODO: Replace with a tree object list
-          treeIds: { type: new GraphQLList(GraphQLString)},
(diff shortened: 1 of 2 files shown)


server/lib/schema/index.ts
```diff
@@ -2094,6 +2094,64 @@ const schema = new GraphQLSchema({
           },
         ),
       },
+      updateTreeExpertAssignments: {
```
zjmiller: In the Mosaic codebase I'd prefer for the concept "assignment" to uniformly mean the assignment of a user to a workspace during an experiment. So I'd prefer we change the name of this mutation to something like `updateTreeExperts`.

server/lib/schema/index.ts
```diff
@@ -2094,6 +2094,64 @@ const schema = new GraphQLSchema({
...
+        type: GraphQLBoolean,
+        args: {
+          treeId: { type: GraphQLString },
+          honestEmails: { type: GraphQLList(GraphQLString) },
```
zjmiller: To maintain consistency with the other API endpoint, we should probably use `emailsOfHonestExperts` and `emailsOfMaliciousExperts` as the names for these arguments.

server/lib/schema/index.ts
```diff
@@ -2094,6 +2094,64 @@ const schema = new GraphQLSchema({
...
+          maliciousEmails: { type: GraphQLList(GraphQLString) },
+        },
+        resolve: requireAdmin(
+          "You must be logged in as an admin to update experiment expert assignments",
```
zjmiller: "update experiment expert assignments" to something like "update the experts for a tree"

server/lib/schema/index.ts
andrewschreiber: Good catch. I think it may be worthwhile to update `mosaic-api-scripts` as well to reduce confusion. Will talk to ben about that.

zjmiller: I'm honestly not too worried about that. I think things are modular enough that I'm happy to defer to Ben R. and Mati about however they want to organize the scripts themselves.

andrewschreiber:
> Have you checked with Ben R. and Mati that what we want for this endpoint is a total reset with respect to a tree's experts? For example, it might be that they plan mostly to add a single expert, and so it will be a lot more work for them to ensure existing experts get re-added every time.
> 
> Also, there's a nice mini-refactoring opportunity here: the `TreeId` and `UserId` on the the `UserTreeOracleRelation` should probably both be lowercase. Maybe you could do a PR after this one where you change those to lowercase?

We didn't discuss it in detail, so let's definitely check in with @brachbach and Mati before merging this. I guess it would be easy to add a bool flag that swaps between deleting or adding to existing expert _designations_. I'll just do that, for better flexibility.

zjmiller:
"Designations" :)

zjmiller:
Cool sounds good, I'm not 100% sure how they anticipate using the API, but we should definitely find out and all sync on exactly what happens.

server/lib/schema/index.ts
andrewschreiber: Noticing that we’re slowly migrating from `Oracle` to `Expert` terminology in the codebase. I’m on the fence about just renaming everything `Expert`, as I feel like code like this (unrelated to PR) could be confusing. 
image
My feeling is that with Mosaic2 on the way, it's probably not worth the time. We should just use Oracle term in the codebase. The exception being for user-facing language.

zjmiller: Yeah I see where you're coming from. Ever since we decided to use "expert" instead of "oracle", I've used "expert" when creating a new variable/field, e.g., `isAwaitingHonestExpertDecision`. I do think there's an argument that I should have just kept using "oracle", though.

Mati-Roy:
What's the plan for merging?

Mati-Roy:
@andrewschreiber , @zjmiller , @brachbach 
Maybe ideally we only want to allow judges becoming experts, or honest experts becoming malicious experts, and have the API only update based on the email addresses we send.

But actually, before updating a tree, I could write a SQL query to extract the current assignment (lmk if you already have such a query). And so a complete overwrite would be fine to.

If you're good with just merging as is, let me know and I'll see to write the query. I need to concatenate all email addresses for a tree in one row; I know it's doable in t-SQL, so it should be in PostgreSQL as well.

server/lib/schema/index.ts
brachbach: I do think this is a good point -- I had been using "assignment" to refer to "we assigned this expert to be on this tree via the API".

The new terminology I've been using (though haven't standardized everywhere, e.g. in mosaic-api-scripts) is "allocation" -- "we allocated this expert to be on this tree via the API"

Definitely taking suggestions for better terminology though :)

brachbach: I do think this is a good point -- I had been using "assignment" to refer to "we assigned this expert to be on this tree via the API".

I like "designate"/"designations" as @andrewschreiber suggested below, will plan to switch to that unless someone else wants to propose something different

brachbach:
Currently the endpoint can only update one tree at a time, and [the associated mosaic-api-script](https://github.com/oughtinc/mosaic-api-scripts/pull/16) also only updates one tree at a time, right?
- if so, I think that's too bad -- I think we'd basically always want to update multiple trees at once (because if it were just one tree, we'd just do it manually)
- so I think we at least want to change the mosaic-api-script to allow updating multiple trees at once. In fact, I think we should change that mosaic-api-scripts PR pretty substantially, I'll submit my comments soon
- it'd also be sort of nice to change this PR to do multiple trees at once, *for the purpose of being [atomic](https://en.wikipedia.org/wiki/ACID#Atomicity)*. There's currently a [bug] (https://github.com/oughtinc/mosaic/issues/723) where API calls to create an experiment or add trees to an experiment are _not_ atomic.
  - I *don't* think it's likely that it's worth changing this PR to allow changing multiple trees at once for *performance* reasons. I think it's fine if it takes an hour to change assignments on 1000 trees (likely the most extreme case)
  - it *might* be worth changing this PR to make it easier to change the mosaic-api-script to allow changing multiple trees at once. Will leave that up to @andrewschreiber

brachbach:
Re: should the endpoint remove all existing expert assignments -- I lean towards *yes*, and no option not to.

My main reason is that I think we (the experiment running team) should maintain spreadsheets that say who is supposed to be assigned to every tree

So, to make reassignments, we'd change the sheet and then feed it through the update script.

@Mati-Roy, do you understand what I mean? Do you agree that it's a good idea? TBC, I'm proposing that we'd do this *instead* of grabbing current assignments with a SQL query

andrewschreiber:
I agree Ben - seems more principled to require an up-to-date set of excel sheets. Sort of like keeping the migrations for a database.

I've written it such that by default, it deletes all previous assignments. However there is a flag that disables the deletion, incase of (hopefully rare) patch fixes. Seems like it addresses both concerns, so I'm in favor of using this existing structure until we think of a better one.

brachbach:
@andrewschreiber , oh, great!

I missed that that optional flag was already there, give that, I agree that we should leave it as is :)

zjmiller:
OK so my understanding is that this PR needs (1) to be set up so as to allow multiple tree updates with one API request, and (2) wrap everything in a transaction.

brachbach:
@zjmiller -- that sounds good, though I'd say 1. is a must-have and 2. is a nice-to-have

andrewschreiber:
> @zjmiller -- that sounds good, though I'd say 1. is a must-have and 2. is a nice-to-have

zjmiller:
Cool I'll review soon.

andrewschreiber:
Ready for review! @zjmiller

server/lib/schema/index.ts
```diff
@@ -2113,6 +2125,98 @@ const schema = new GraphQLSchema({
...
+
+                const tree: Tree = await Tree.findByPk(treeId);
+                if (tree === null) {
+                  throw new Error(`No tree found for TreeId ${treeId}`);
+                }
+
+                if (replaceOracles) {
+                  await UserTreeOracleRelation.destroy({
+                    where: {
+                      TreeId: tree.id,
```
zjmiller: I'll make an issue about refactoring `UserTreeOracleRelation` to use `treeId` and `userId` instead of `TreeId` and `UserID`. No need to do anything this PR.

server/lib/schema/index.ts
```diff
@@ -2113,6 +2125,98 @@ const schema = new GraphQLSchema({
...
+                    },
+                    transaction,
+                  });
+                }
+
+                for (const email of emailsOfHonestOracles) {
+                  const user: User | null = await User.findOne({
+                    where: {
+                      email: email.toLowerCase(),
```
zjmiller: I'll make an issue about using something other than `String.prototype.toLowerCase` here. Maybe `Sequelize.Op.iLike` which is based on Postgres `ILIKE`? Again, no need to do anything this PR.

server/lib/schema/index.ts
```diff
@@ -2113,6 +2125,98 @@ const schema = new GraphQLSchema({
...
+
+                  oracleRelation.update({ isMalicious: true }, transaction);
+                }
+              }
+              await transaction.commit();
+              return true;
+            } catch (err) {
+              console.error("updateTreeExperts error", err);
+              await transaction.rollback();
+              return false;
```
zjmiller: I think we want to throw an error here instead of `console.error`. If the error is thrown, then the person who ran the script will see the error message. `console.error` will just keep it on the server right?

server/lib/schema/index.ts
```diff
@@ -396,6 +398,16 @@ const ExperimentInput = new GraphQLInputObjectType({
   },
 });
 
+const TreeReplaceExpertInput = new GraphQLInputObjectType({
+  name: "treeReplaceExpertInput",
+  fields: {
+    treeId: { type: GraphQLString },
+    replaceOracles: { type: GraphQLBoolean },
```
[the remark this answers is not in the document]
andrewschreiber: Good idea.

server/lib/schema/index.ts
andrewschreiber: Good point. The downstream consumer will get an error, but it won't include this text.

brachbach:
@zjmiller 👍.

@andrewschreiber 's companion `mosaic-api-scripts` PR (https://github.com/oughtinc/mosaic-api-scripts/pull/16) tests it out a bit, but would be nice for @Mati-Roy to do a test run of an actual  use of this functionality before we merge

Mati-Roy:
@brachbach 

re 2: done; see: https://ought-inc.slack.com/archives/CL6PLP82J/p1568874834040500 and https://docs.google.com/spreadsheets/u/1/d/1q_DAP_obxGnKHEKiImeVdPNEm5HPrPwcmsfH3F8mk0M/edit#gid=0

re 1: @zjmiller , what's the easiest way to extract designations? should that be done through the API?

zjmiller:
Yeah if you can access the GraphQL API that's the best way.

The following can be nested side inside the `tree` part of a GraphQL query. We can chat about the details on Slack because it's potentially a bit involved.

```
oracleRelations {
  isMalicious
  user {
    email
  }
}
```

Mati-Roy:
> it's potentially a bit involved.

@brachbach , then I wouldn't do it (at least not now) b/c it doesn't seem a priority. we can just use the Google Sheet. what do you think?

brachbach:
yeah I've always been happy with "Google Sheet as source of truth" :)

Mati-Roy:
good ^_^

PR #725, opened 2019-09-04
Allow experts to skip workspaces

zjmiller:


andrewschreiber:
Instead of it's current, spot in the top center. I'm thinking the skip button could be in the `Answer Candidate` box, right aligned?

andrewschreiber:
Current and target spot (black box). GitHub image uploading is down, so I had to use this: https://ibb.co/xDJpgkh

andrewschreiber:
@brachbach @Mati-Roy Any UI feedback here?

Mati-Roy:
I don't feel strongly one way or another.

PR #733, opened 2019-09-09
Per tree depth limits

zjmiller:


zjmiller:
When creating a tree add a field `depthLimit: 3` to override the default depth limit of 6.

brachbach:
cool -- for updating, can you choose which things to update?

i.e. if I only include `depthLimit: n`, then only depth limit gets updated, and if I only include new expert assignments, then only expert assignments get updated?

Oh wait or is it a separate endpoint?

zjmiller:
Andrew's "allow tree updates via the API" only allows the changing of experts. I assume we should ask him to include `depthLimit` as well, and `schedulingPriority`?

brachbach:
ah, I missed that it only changed experts.

Hmm.. I actually don't feel strongly about being able to update `depthLimit` and `schedulingPriority` via the API

I think we're less likely to want to change those ones trees are created than we are to want to change expert assignments

(@Mati-Roy, correct me if that seems wrong)

I'd rather see Andrew just try to get that PR merged, rather than add stuff to it

brachbach:
Created a new issue for being able to change `schedulingPriority` via the API: https://github.com/oughtinc/mosaic/issues/735

client/src/pages/RootWorkspacePage/AdminControls/index.tsx
```diff
@@ -158,6 +170,68 @@ class AdminControlsPresentational extends React.Component<any, any> {
...
+            }}
+          >
+            Error: non-numeric value entered
+          </div>
+          <FormControl
+            onChange={e => {
+              const target = e.target as HTMLInputElement;
+
+              this.setState({
+                depthLimit: target.value,
```
andrewschreiber: Why not set this as a Number instead of wrapping later values?

zjmiller: I tried that initially but then if the user inputs a non-numeric value into the field, the UI breaks. As a controlled input, we want this bit of state to be able to equal any possible value (and then warn the user if they've entered a non-numeric value and give them a chance to correct it). 

If you can get a less awkward approach to work, let me know. Though probably not high priority right now.

andrewschreiber: I think a less awkward approach would be to restrict the user's inputs to valid ones instead of throwing an error message. This also prevents NaN cases, simplifying the logic. Here is a regex I found (but did not test) on SO.
```
   onChange(e){
      const target = e.target as HTMLInputElement;
      const regex = /^[0-9\b]+$/;
      if (target.value === '' || regex.test(target.value)) {
         this.setState({ depthLimit: Number(target.value)})
      }
   }
```
Another thing we need to be careful with is decimals, as depthLimit should always be an Int.

andrewschreiber: The only really important thing for me though is to make sure we only accept `Int`, to prevent a graphql error. Did you try writing `2.4` or something as a depthLimit?

zjmiller:
I just tested out the API at the same URL and it correctly set the depth limit to 3. So my current best guess is that the script isn't quite right (or is bringing up some edge case the API can't process) and the default value of 6 is getting used.

Can you link me to the script you're using? Also, if there's some way for the script to log the object it's sending to the API, that would be useful to look at.

Mati-Roy:
> Can you link me to the script you're using?

https://github.com/oughtinc/mosaic-api-scripts/pull/13/commits/587a5cb139e9dc95120e08541d8cb7a66ce75a4c

> Also, if there's some way for the script to log the object it's sending to the API, that would be useful to look at.

I'll send you the json file in DM

PR #738, opened 2019-09-14
Fix issues with "update experts on existing trees" API endpoint

zjmiller:


zjmiller:
Made sure transaction always got properly passed, and added a few missing `await`s.

server/lib/schema/index.ts
```diff
@@ -2189,7 +2189,7 @@ const schema = new GraphQLSchema({
                     );
                   }
 
-                  tree.$add("oracle", user, transaction);
+                  await tree.$add("oracle", user, { transaction });
```
andrewschreiber: I wonder if there was a race condition here previously that was causing the inconsistent behavior.

server/lib/schema/index.ts
```diff
@@ -2204,14 +2204,16 @@ const schema = new GraphQLSchema({
...
 
                   const oracleRelation: UserTreeOracleRelation | null = await UserTreeOracleRelation.findOne(
                     {
                       where: {
                         UserId: user.id,
                         TreeId: tree.id,
                       },
+
+                      transaction,
```
andrewschreiber: Why pass the transaction here?

server/lib/schema/index.ts
andrewschreiber: Good catch on wrapping the transaction with an object.

server/lib/schema/index.ts
andrewschreiber: It looks like the field expects a `FindOptions` object, of which `transaction` is one property.  (Specifically, `FindOptions` is an interface that extends `QueryOptions`, which extends `Transactionable`, which has a `transaction` property). So I'd expect this to need to be surrounded by curly braces as well.

andrewschreiber: Somewhat disappointed TypeScript didn't catch the `transaction type errors I made in the past PR. I wonder if there is a stricter check we could use. It shouldn't have accepted a `Transaction` object when it was looking for a `FindOptions` object...

PR #744, opened 2019-09-17
Hide archive button for non-root workspaces

### LINKED ISSUE #737: Remove the ability to archive subquestions
In vanilla FE, archiving is sort of confusing. IMO the main point of it is if you realize that you made an error in your subquestion and want to get rid of it before it gets assigned to anyone. That is, I don't think it makes much sense to archive a subquestion that's already received an answer.

In MIB, your judge assignment ends as soon as you ask a subquestion, and the workspace won't get assigned to someone else until the subquestion produces an answer. So, there's no good case for using archiving.

Given that it adds complication to the game and is generally confusing, I think we should just get rid of it (at least for now).

We could wait until we've fully rolled out MIB, but I don't think that's necessary because I don't think that archiving is very useful even in vanilla FE.

I'm guessing that removing the ability to archive would be extremely easy, since we can just remove the UI element. I think we probably don't want to put in the effort to remove all code related to archiving, especially since there's some chance we'll want to put it back in later.

andrewschreiber:
closes #737

### DIFF
--- a/client/src/components/NewBlockForm.tsx
+++ b/client/src/components/NewBlockForm.tsx
@@ -65,7 +65,7 @@ export class NewBlockFormPresentational extends React.Component<any, any> {
     };
   }
 
-  public componentWillReceiveProps(newProps) {
+  public componentWillReceiveProps(newProps: any) {
     if (
       this.state.pending &&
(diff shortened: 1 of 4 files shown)


andrewschreiber:
Tested. Archive is hidden when appropriate.

Top-level view:
image

Non-top level judge view:
image

PR #750, opened 2019-09-24
Judge-specific text display

### LINKED ISSUE #736: Be able to show arbitrary text (including hyperlinks) in every workspace of a tree
The current primary motivation for this is to be able to link to a doc that specifies what strategy judges should use on the tree.

But we may want it for other things in the future.

We could do this as an extension of https://github.com/oughtinc/mosaic/issues/626 -- I'd be totally fine with something hacky like:
1. All text in the root-level stratchpad gets shown to experts
2. All text between, say < and > also gets shown to judges

andrewschreiber:
closes #736

Wrap scratchpad text in `<judge> judge text here </judge>` for judge-specific text. Experts will be able to see the judge text, but judges wont be able to see anything outside the judge text.

### DIFF
--- a/client/src/pages/EpisodeShowPage/index.tsx
+++ b/client/src/pages/EpisodeShowPage/index.tsx
@@ -683,11 +683,11 @@ export class WorkspaceView extends React.Component<any, any> {
       !isOracleWorkspace &&
       doesRootWorkspaceScratchpadContainRelevantContent;
 
-    const judgeRegex = /(?<=<judge>)(.*)(?=<judge>)/;
+    const judgeRegex = /<judge>(.*)<\/judge>/;
     const judgeTextArray = judgeRegex.exec(
       listOfSlateNodesToText(_.get(rootWorkspaceScratchPad, "value"), false),
     );
-    const judgeText = judgeTextArray ? judgeTextArray[0] : null;
+    const judgeText = judgeTextArray ? judgeTextArray[1] : null;
 
     return (
       <div>
--- a/client/src/pages/EpisodeShowPage/index.tsx
+++ b/client/src/pages/EpisodeShowPage/index.tsx
@@ -48,6 +48,7 @@ import {
 
 import { ExpandAllPointersBtn } from "./ExpandAllPointersBtn";
 import { databaseJSONToValue } from "../../lib/slateParser";
+import { listOfSlateNodesToText } from "../../lib/slateParser";
 
 import {
   ORACLE_MODE_QUERY,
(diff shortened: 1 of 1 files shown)


andrewschreiber:
Tested and looking as expected.

Expert View:
image

Judge View:
image

andrewschreiber:
Realizing we should probably document this new functionality somewhere. Thoughts? @brachbach @Mati-Roy

brachbach:
1. I'd do `<judge>...</judge>` instead of `<judge>...<judge>`, for consistency with HTML
2. I think would make sense to document in the `mosaic-api-scripts` README under the section about creating experiments/trees (since we mainly expect to use the functionality via the API)

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -677,6 +678,17 @@ export class WorkspaceView extends React.Component<any, any> {
       ? this.getAvailablePointers(workspace, rootWorkspaceScratchPad)
       : this.getAvailablePointers(workspace);
 
+    const shouldShowJudgeTextFromScratchpad =
+      hasParent &&
+      !isOracleWorkspace &&
+      doesRootWorkspaceScratchpadContainRelevantContent;
+
+    const judgeRegex = /(?<=<judge>)(.*)(?=<judge>)/;
```
zjmiller: Lookbehind `?<=` is not supported by Firefox, so Firefox crashes when loading a workspace.

I'm pretty sure you can change this regex to `/<judge>(.*)<judge>/`. And then the second item in the result array returned by `.exec` will be the match within the parentheses (i.e., the text in between the tags). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -677,6 +678,17 @@ export class WorkspaceView extends React.Component<any, any> {
...
+    const judgeTextArray = judgeRegex.exec(
+      listOfSlateNodesToText(_.get(rootWorkspaceScratchPad, "value"), false),
+    );
+    const judgeText = judgeTextArray ? judgeTextArray[0] : null;
```
zjmiller: So if you change the regex as mentioned above, the `judgeTextArray[0]` should be `judgeTextArray[1]`.

Mati-Roy:
what use case did we develop this functionality for?

andrewschreiber:
The use case is inline communication of specific strategies that we want the participants to use, according to @brachbach

client/src/pages/EpisodeShowPage/index.tsx
andrewschreiber: Good call, will update. That regex would work, according to https://regexr.com/

andrewschreiber:
Updated regex to work on Firefox and terminate with </judge>
image

Will make a mosaic-api-scripts PR with the documentation after this is merged.

Ready for 2nd review pass @zjmiller

client/src/pages/EpisodeShowPage/index.tsx
```diff
@@ -683,11 +683,11 @@ export class WorkspaceView extends React.Component<any, any> {
       !isOracleWorkspace &&
       doesRootWorkspaceScratchpadContainRelevantContent;
 
-    const judgeRegex = /(?<=<judge>)(.*)(?=<judge>)/;
+    const judgeRegex = /<judge>(.*)<\/judge>/;
```
zjmiller: Wait, do we want `<judge><judge>` or `<judge></judge>`?

When I first reviewed it was the former, but now it's the latter.

andrewschreiber: The later - per Ben's request / reference to HTML

PR #763, opened 2019-10-09
Scheduling performance improvements

### LINKED ISSUE #743: The app seems to be getting quite slow for participants
Erol Can Akbaba Yesterday at 1:39 PM
This isn't really a bug, but performance is starting to become an issue for me. The load time for listing of previous workspaces is quite long for some time but that doesn't impact my work. Load times for finding a workspace is getting longer and longer - sometimes I refresh the page after a minute or so to get it going.

:+1:
5

https://ought-inc.slack.com/archives/CMPKU3B5F/p1568666376008000

andrewschreiber:
Refactors scheduling logic to combine queries into bigger batches and remove redundant/outdated computation.

closes #743

### DIFF
--- a/client/src/pages/NextEpisodeShowPage.tsx
+++ b/client/src/pages/NextEpisodeShowPage.tsx
@@ -161,7 +161,6 @@ export class NextEpisodeShowPagePresentational extends React.Component<
     if (this.state.refreshCountdown === 0) {
       location.reload();
     }
-    console.log("Oracle Scheduling failed?", this.state.oracleSchedulingFailed);
 
     if (this.state.oracleSchedulingErrored) {
       return (
--- a/client/src/pages/NextEpisodeShowPage.tsx
+++ b/client/src/pages/NextEpisodeShowPage.tsx
@@ -81,8 +81,9 @@ export class NextEpisodeShowPagePresentational extends React.Component<
   public constructor(props: any) {
     super(props);
     this.state = {
-      normalSchedulingFailed: false,
       oracleSchedulingFailed: false,
+      oracleSchedulingErrored: false,
+      oracleSchedulingErrorMessage: null,
       refreshCountdown: 10,
       workspaceId: undefined,
       isCountingDown: false,
@@ -95,7 +96,10 @@ export class NextEpisodeShowPagePresentational extends React.Component<
   }
 
   public async componentDidMount() {
-    let response, normalSchedulingFailed, oracleSchedulingFailed;
+    let response,
+      oracleSchedulingFailed,
+      oracleSchedulingErrored,
+      oracleSchedulingErrorMessage;
 
     const experimentId = getExperimentIdOrSerialIdFromQueryParams(
       window.location.search,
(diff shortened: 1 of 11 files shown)


andrewschreiber:
PR status: most targeted changes have been made. Need to clean up code and test on a copy of prod DB.

zjmiller:
As mentioned on Slack, when I deployed this PR to staging, attempting to find a workspace in experiment "L2 MIB Experiment" resulted in the following console error:

image

andrewschreiber:
Fixed a few issues that I found. Seemingly including `getChildWorkspaces` error, but I wasn't able to consistently reproduce it in the first place.

Now deployed on https://mosaic-staging-1.herokuapp.com with a copy of prod database.

Did some casual performance testing by making myself an honest expert and malicious expert on some trees.

Looking for next workspace seems to be near-instant on `staging`, compared to ~8 seconds on `production`. Both are under low/zero load right now. Looking forward to more load testing tomorrow.

client/src/pages/NextEpisodeShowPage.tsx
```diff
@@ -146,49 +161,35 @@ export class NextEpisodeShowPagePresentational extends React.Component<
     if (this.state.refreshCountdown === 0) {
       location.reload();
     }
+    console.log("Oracle Scheduling failed?", this.state.oracleSchedulingFailed);
```
zjmiller: Take this out?

server/lib/scheduler/Scheduler.ts
```diff
@@ -122,102 +127,67 @@ class Scheduler {
...
-        );
-
-        return assignedWorkspaceId;
-      }
-
-      throw new Error("No eligible workspace");
+    // Should always be in oracle mode
+    if (!this.isInOracleMode.getValue()) {
+      throw new Error(
+        "Not in Oracle Mode. If you see this, please contact us. Then try logging out and logging back in.",
```
zjmiller: I would maybe take out the "Then try logging out and logging back in". Not a blocking issue, though. If they see this error it's the least of our worries.

server/lib/scheduler/Scheduler.ts
```diff
@@ -122,102 +127,67 @@ class Scheduler {
...
-    const assignedWorkspace = pickRandomItemFromArray(actionableWorkspaces);
-    const isThisAssignmentTimed = await assignedWorkspace.hasTimeBudgetOfRootParent;
+    const actionableWorkspace = await this.getActionableWorkspaceInOracleMode(
+      userId,
+    );
 
-    // This is intended to mitigate a race condition where two users
-    // are both looking for a new assignment at the same time and
-    // get assigned to the same workspace due to async code in the scheduler
-    if (this.schedule.isWorkspaceCurrentlyBeingWorkedOn(assignedWorkspace)) {
```
zjmiller: I'd put this back. It's important this check happens as close to the end as possible or else it won't mitigate the race condition aspect of the problem.

server/lib/scheduler/Scheduler.ts
```diff
@@ -261,126 +232,118 @@ class Scheduler {
...
+        const workspacesInTree: Workspace[] = await this.fetchAllWorkspacesInTree(
+          randomlySelectedRootWorkspace,
         );
 
-        const workspacesToConsider = await this.filterByWhetherCurrentlyBeingWorkedOn(
+        const oracleEligibleWorkspaces: Workspace[] = isMaliciousInTree
+          ? this.filterByWhetherEligibleForMaliciousOracle(workspacesInTree)
+          : this.filterByWhetherEligibleForHonestOracle(workspacesInTree);
+
+        const staleOracleEligibleWorkspaces = this.filterByStaleness(
```
zjmiller: It doesn't look like `staleOracleEligibleWorkspaces` is being used. We do need to make sure to only ever assign a stale workspace though.

andrewschreiber: Good catch!

server/lib/scheduler/Scheduler.ts
andrewschreiber: Yeah fair point. Can't be too safe here, even if it's some extra compute.

server/lib/scheduler/Scheduler.ts
```diff
@@ -251,24 +222,19 @@ class Scheduler {
...
+        const randomlySelectedRootWorkspace = this.getRandomTreeRootWorkspaceWorkedOnLeastRecentlyByUser(
           userId,
-          rootWorkspacesOfTrees,
+          oracleTreesToConsider,
         );
-        const randomlySelectedRootWorkspace = pickRandomItemFromArray(
-          leastRecentlyWorkedOnTreeRootWorkspaces,
-        );
-
+        if (!randomlySelectedRootWorkspace) {
```
zjmiller: Maybe throw an error here? This condition should never be true so if it is something went wrong. Same thing with line 285.

andrewschreiber: On second look, definitely agreed. We'd want to know if this is happening, instead of people just getting fewer results.

PR #765, opened 2019-10-12
Fix request errors on workspace & findNextWorkspace

### LINKED ISSUE #764: Fix server failed response rate
Although #763 reduced the error rate substantially - as seem by before/after Friday - there are still a substantial amount of timeout errors.
image

It seems our timeout is ~30 seconds. This seems like a good upper bound: https://ux.stackexchange.com/questions/45323/whats-reasonable-time-lag-before-displaying-a-timeout-message.

Heroku gives us db-level analytics via `heroku pg:outliners -a mosaic-prod` on which queries are the most expensive:
image
More info is available here: https://data.heroku.com/datastores/b2ed9b32-2f37-4563-a249-e580d25e5fda#diagnose

andrewschreiber:
closes #764

### DIFF
--- a/server/lib/models/workspace.ts
+++ b/server/lib/models/workspace.ts
@@ -209,6 +209,8 @@ export default class Workspace extends Model<Workspace> {
       // This pointer is contained in the malicious question.
       const maliciousWorkspace = await Workspace.findByPk(workspace.parentId);
 
+      // TODO: How to handle when workspace can be null
+
       const maliciousQuestionBlock = (await maliciousWorkspace.$get(
         "blocks",
       )).find(b => b.type === "QUESTION");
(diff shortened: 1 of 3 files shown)


server/lib/models/workspace.ts
```diff
@@ -208,6 +208,11 @@ export default class Workspace extends Model<Workspace> {
       // not the pointer containing the honest answer candidate.
       // This pointer is contained in the malicious question.
       const maliciousWorkspace = await Workspace.findByPk(workspace.parentId);
+      if (!maliciousWorkspace) {
```
zjmiller: I would chance this condition to `!maliciousWorkspace || !maliciousWorkspace.isEligibleForMaliciousOracle)`. So now even if a workspace has a parent, if this parent isn't malicious then we stop here.

server/lib/models/workspace.ts
```diff
@@ -250,6 +256,9 @@ export default class Workspace extends Model<Workspace> {
       // contains the honest answer candidate, and this first appears in the
       // malicious question.
       const maliciousWorkspace = await Workspace.findByPk(workspace.parentId);
+      if (!maliciousWorkspace) {
```
zjmiller: Same as above.

